PR #2, opened 2024-05-09
optimize mmr logic

sophiasun0515:
Following method 1 as we discussed, moved the similarity matrix logic outside of `generate_recommendations()` to avoid repeated computation for each user. 

Note: did not test yet.

### DIFF
--- a/src/poprox_recommender/nrms.py
+++ b/src/poprox_recommender/nrms.py
@@ -162,6 +162,8 @@ def generate_recommendations(model, articles, article_vectors, similarity_matrix
 
 
 def select_with_model(
+    todays_articles,
+    todays_article_vectors,
     article_similarity_matrix,
     past_article_features,
     clicked_articles,
@@ -179,12 +181,26 @@ def select_with_model(
 
     recommendations = generate_recommendations(
         model,
+        todays_articles, 
+        todays_article_vectors,
         article_similarity_matrix,
         user_embeddings,
         num_slots=num_slots,
     )
 
     return recommendations
+
+
+def compute_similarity_matrix(todays_article_vectors):
+    num_values = len(todays_article_vectors)
+    similarity_matrix = np.zeros((num_values, num_values))
+    for i, value1 in tqdm(enumerate(todays_article_vectors), total=num_values):
+        value1 = value1.detach().cpu()
+        for j, value2 in enumerate(todays_article_vectors):
+            if i <= j:
+                value2 = value2.detach().cpu()
+                similarity_matrix[i, j] = similarity_matrix[j, i] = np.dot(value1, value2)
+    return similarity_matrix
 
 
 def select_articles(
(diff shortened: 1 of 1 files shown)


src/poprox_recommender/nrms.py
```diff
@@ -172,29 +162,24 @@ def generate_recommendations(model, articles, article_vectors, user_embeddings,
...
+    # Build embedding tables
     past_article_embeddings, _ = build_article_embeddings(past_article_features, model, model_device)
     user_embeddings = build_user_embeddings(user_df, past_article_embeddings, model, model_device, max_clicks_per_user)
 
-    _, todays_article_vectors = build_article_embeddings(todays_article_features, model, model_device)
-
     recommendations = generate_recommendations(
         model,
-        todays_articles,
-        todays_article_vectors,
```
karlhigley, line 197: It looks like `generate_recommendations` still uses `articles` and `article_vectors` so we should keep these

src/poprox_recommender/nrms.py
```diff
@@ -230,15 +215,30 @@ def select_articles(
         token_mapping,
     )
 
+    # Load model
+    model = load_model(model_checkpoint, model_device)
+
+    # Compute today's article similarity matrix
```
karlhigley, line 221: This chunk of code that computes the article similarity matrix would make a good function. Maybe something like `compute_similarity_matrix()`?

src/poprox_recommender/nrms.py
sophiasun0515: Thanks for catching that! I know there must be something missing 😅Hopefully when we have more test cases built up it'll be easier to catch!

src/poprox_recommender/nrms.py
sophiasun0515: Make sense, refactored!

src/poprox_recommender/nrms.py
karlhigley: 💯 

Every change is like walking a high wire right now, and changing that should be a high priority. Talked to @mdekstrand about that a bit this afternoon, and we were thinking one of the things this repo should have is a script that downloads the model and some data, runs them through the recommender, and just prints out the results. It's not really a substitute for a test suite, but it would give us a way to double-check that things are working the way we think they should by 👀 ing the output

karlhigley:
It looks good to me! I'll give it a test run to shake out any further issues and merge if that goes smoothly

src/poprox_recommender/nrms.py
sophiasun0515: Totally agree. For the model / test data downloading part I assume we might need to integrate the S3 logic? Happy to help with that if needed (or if it's already in your next week's plan then I can wait until you create a PR and add more stuff to that).

mdekstrand: We can probably just have DVC pull directly from S3: https://dvc.org/doc/command-reference/import-url

Then we can load from that local file (and speed up iteration).

mdekstrand: (Adding DVC to this repo is a "do this very soon" task, and I was going to connect with @aisa6148 on Monday about kicking that off.)

PR #3, opened 2024-05-13
Create an API endpoint as a Serverless Lambda function

karlhigley:
This sets up a simulated Lambda endpoint for local development that should run without AWS credentials. Once we're happy with how the code in this repo integrates with the real Lambda endpoints in `poprox-core` (via direct package dependency), we can configure this to be deployed as a real HTTP endpoint that `poprox-core` makes calls to.


src/poprox_recommender/default.py
```diff
@@ -146,7 +146,8 @@ def mmr_diversification(rewards, similarity_matrix, theta, topk):
             if mr_i >= best_MR and i not in S:
                 best_MR = mr_i
                 candidate = i
-        S.append(candidate)
+        if candidate:
+            S.append(candidate)
```
karlhigley, line 150: This is a tiny fix for an issue that AFAIK only comes up when you try to recommend from a a very small pool of items (like 1 item 😅 .) When there aren't any more items to add to the list of recommendations, the `candidate` was still getting appended even if it was `None`

src/poprox_recommender/paths.py
```diff
@@ -0,0 +1,5 @@
+from pathlib import Path
+
+
+def project_root() -> Path:
+    return Path(__file__).parent.parent.parent
```
karlhigley, line 5: This little utility function helps us find the root path of the repo, which makes it easier to find the `models` folder we have set up with DVC as the destination for downloaded model checkpoints/artifacts

tests/basic-request.json
```diff
@@ -0,0 +1,24 @@
...
+            "url": "url 2"
+        }
+    ],
+    "click_data": {
+        "user 1": [
+            "url 1"
+        ]
+    },
+    "num_recs": 1
+}
```
karlhigley, line 24: This is just test data to be used with the command included in the README that sends it to the endpoint with `curl`

src/poprox_recommender/handler.py
```diff
@@ -0,0 +1,67 @@
...
+        token_mapping = dict(pd.read_table(f, na_filter=False).values.tolist())
+
+    return token_mapping
+
+
+MODEL, DEVICE = load_model()
+TOKEN_MAPPING = load_token_mapping()
+
+
+def generate_recs(event, context):
```
karlhigley, line 44: This is the Lambda handler function that gets called when HTTP requests are received. It grabs a few parameters from the body of the request, feeds them into `select_articles` (where all the magic happens), and sends the output back as a response

serverless.yml
```diff
@@ -0,0 +1,50 @@
+service: poprox-default-recommender
```
karlhigley, line 1: This configuration tells `serverless` how to set our code up as an AWS Lambda function. Most of it is boilerplate, but the interesting part is the `functions:` section at the bottom

serverless.yml
```diff
@@ -0,0 +1,50 @@
...
+package:
+  patterns:
+    - "!venv/**"
+    - "!models/**"
+    - "!node_modules/**"
+    - "!tests/**"
+    - "**.py"
+    - "**.html"
+
+functions:
```
karlhigley, line 44: This is where we link incoming request URLs/paths to specific Python functions that should run to process the request. `handler` is just the import path of the function, and `httpAPI: path` is where we set the URL. It could be something like `/generate` to be explicit, but I think there's only one URL and one function needed so maybe `/` is okay?

sophiasun0515:
Sorry -- I accidentally added the test files to this PR as well, but it's just a small change so hopefully not affect whatever this PR aims for 😅

sophiasun0515:
lgtm!

PR #5, opened 2024-05-15
Use Pydantic models to represent domain concepts, starting with `Article`

karlhigley:
This is a first step toward representing domain concepts like articles, accounts, entities, etc with Python classes instead of dictionaries. That will make it easier for us to validate the data we're working with and know what fields to expect will be present. We'll also get better type-checking since we'll be able to annotate more of the function parameters and return types.

### DIFF
--- a/src/poprox_recommender/default.py
+++ b/src/poprox_recommender/default.py
@@ -1,6 +1,7 @@
 import random
 from dataclasses import dataclass
 from functools import partial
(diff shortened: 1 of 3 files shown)


kluver (approved):
I haven't been keeping up with the NRMS code -- it wouldn't hurt to get a pass from someone who knows that code well.

PR #6, opened 2024-05-15
Environment cleanup, docs, and conda lock

mdekstrand:
This PR does a few cleanup tasks for the repo from what we had with `hatch init` defaults:

[SPEC0]: https://scientific-python.org/specs/spec-0000/

1. Remove classifiers since we aren't publishing to PyPI.
2. Bump the minimum Python version. I used 3.10, as it's the oldest supported under [SPEC0], but I would support requiring 3.11 (or 3.12, but not all packages are necessarily compatible with that yet).
3. Remove the hatch matrix stuff — I don't think testing on dependency matrices is likely to be a priority for our repo, we can pick a Python and use it.
4. Remove mypy and Black in favor of ruff and pyright, and remove configurations for linters we aren't using (e.g. flake8).
5. Specify minimum (and maximum) dependency versions to speed up dependency solving and reduce the risk of surprises.
6. Add metadata to `pyproject.toml` to allow `conda-lock` to create a lockfile from it. Python version is locked at 3.11.
7. Generate and commit a `conda-lock.yml` with locked dependency versions for Linux and macOS (Apple Silicon only right now).
8. Update the README to provide real instructions.
9. Add an `.editorconfig` file.

Open questions:
- is this the right python version requirements?
- are their other dependency version specs we should add?
- do we need to support working with this repo on Windows or Intel Macs?
- what's the story on that `ruff` configuration, do we understand it and like it? (this can be addressed in later changes)


karlhigley:
Looks good! On the open questions:
* I think we're using Python 3.11 when we deploy Lambdas to AWS but I don't know of anything we're using that isn't compatible with 3.10
* I can't think of any further dependency specifiers to add
* Not sure what people on our team are running, but I suppose we'll find out how urgent that support is as people try to use the repo. We probably will want to support as broad a set of architectures as we can since this repo will be what experimenters interact with.
* The ruff config is Hatch defaults. Happy to adjust or have it adjusted but not super-worried about it for this PR.

karlhigley (approved):

mdekstrand:
Agree on broad set, we probably do want particularly to test Windows support before opening the floodgates to others. Linux ARM is low-priority unless either we decide to start deploying on ARM (may be cheaper?), or someone needs it.

(Me, actually — that someone might be me, once the Grace Hopper nodes get installed in our AI cluster.)

PR #7, opened 2024-05-16
Add MIND Test Data And Create Test Case

sophiasun0515:
Changes included in this PR:

1. Updated README file with offline test and dvc data upload steps.
2. Upload MIND Large test data via dvc.
3. Add `eval_with_mind_large_test.py` file that contains offline test for new model development.

- Test result with current model checkpoint:
  `AUC: 0.5140, MRR: 0.2180, nDCG@5: 0.2266, nDCG@10: 0.2910`

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,3 +106,28 @@ $ curl -X POST -H "Content-Type: application/json" -d @tests/basic-request.json 
 
 {"recommendations": {"user 1": [{"article_id": "2", "title": "title 2", "content": "content 2", "url": "url 2"}]}}
 ```
+
+To add large data or model file via dvc:
+
+1. install dvc dependencies if you don't have them already
+
+    `pip install dvc`
+    `pip install dvc_s3`
+
+2. Obtain the credentials for the S3 bucket and put them in `.env` (the environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`).
+If this step does not work, you can also try to directly export them with 
+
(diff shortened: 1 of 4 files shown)


karlhigley:
Added @mdekstrand and @aisa6148 as reviewers for visibility across the team, since I think this may be related to where they want to go with `dvc repro`

mdekstrand (commented):
I've left a few comments, key short-term things:

- can we rely on the dependency specifications (either pyproject.toml or conda-lock.yml) instead of encouraging people to directly install packages?
- can we use existing, tested metric implementations instead of including new hand-rolled ones?

When it comes to actual DVC integration, which looks like a later PR, here's what I usually do (and recommend unless we have a specific reason to do otherwise):

1. Batch-run script that takes the data, models, and test users, and writes the recommendations to a Parquet (or JSON, depending on how we want to structure outputs) file.
2. Evaluation file reads in the recommendations, computes the metrics, and reports them (sometimes with charts — often my eval file is actually a notebook + script pair managed with `jupytext`, which can then integrate into the DVC pipeline). This file should also write the metrics to a JSON file that is tracked in DVC as a `metric` output, so that we can use DVC's metrics support to compare metrics across runs, branches, etc.

sophiasun0515:
This can be closed since the logic is already added to https://github.com/CCRI-POPROX/poprox-recommender/blob/feature/bert-tokenizer-integration/src/poprox_recommender/test_json.py

PR #8, opened 2024-05-17
Create and apply `Account` and `ClickHistory` classes

karlhigley:


src/poprox_recommender/domain/account.py
```diff
@@ -0,0 +1,8 @@
+from uuid import UUID
+
+from pydantic import BaseModel
+
+
+class Account(BaseModel):
+    account_id: UUID = None
+    email: str
```
karlhigley, line 8: This isn't actually used in the recommender code but is used in the associated platform code in `poprox-core`. I'm putting it here anyway so that all the domain object models are in the same package, which will make it easier to extract them from these repos into a separate repo/package in the future

src/poprox_recommender/default.py
```diff
@@ -241,20 +241,21 @@ def select_articles(
...
+    for click_history in click_histories:
+        account_id = click_history.account_id
+        if model_checkpoint and token_mapping and click_history.article_ids:
             user_recs = select_with_model(
                 todays_articles,
                 todays_article_vectors,
                 similarity_matrix,
                 past_article_features,
-                {user: clicks},
+                click_history,
```
kluver, line 252: If I'm reading this and CCRI-POPROX/poprox-core#41 correctly, the select_with_model function is now officially taking an article ID instead of a URL? So long as the recommender code is good with that I don't think we've got a problem here.

kluver (approved):
one minor detail I want to check in on, see comment. Otherwise it all looks groovy.

src/poprox_recommender/default.py
karlhigley: Yeah, that's where I'm trying to take it. The recommender code doesn't actually care what identifier is used as long it's a string, and the URL itself is currently only used by the mailer

PR #9, opened 2024-05-17
Set up deployment with a Docker container but local execution without

karlhigley:
This sets up the basic mechanism for switching between running the Lambda handler function directly with simulated AWS support locally and running the handler in a Docker container with real AWS infrastructure when deployed. It doesn't yet configure anything about the HTTP endpoint that gets deployed (e.g. VPCs.)


pyproject.toml
```diff
@@ -17,6 +17,9 @@ Documentation = "https://github.com/CCRI-POPROX/poprox-recommender#readme"
 Issues = "https://github.com/CCRI-POPROX/poprox-recommender/issues"
 Source = "https://github.com/CCRI-POPROX/poprox-recommender"
 
+[tool.hatch.build.targets.wheel.force-include]
+"src/models" = "models/"
```
karlhigley, line 21: The model artifacts are baked into the `whl` that gets installed into the Docker container for deployment

src/poprox_recommender/paths.py
```diff
@@ -2,4 +2,4 @@
 
 
 def project_root() -> Path:
-    return Path(__file__).parent.parent.parent
+    return Path(__file__).parent.parent
```
karlhigley, line 5: This adjusts for models moving from `models/` to `src/models`, which changed in order to bake the models into the `whl`

serverless.yml
```diff
@@ -19,10 +20,23 @@ provider:
...
   - serverless-offline
   - serverless-python-requirements
 custom:
+  "offline:true": # offline enabled parameters
+    enabled: true
+    filePrefix: "offline."
+  "offline:false": # offline disabled parameters
+    enabled: false
+    filePrefix: ""
+  offline: ${self:custom.offline:${strToBool(${env:IS_OFFLINE, false})}}
```
karlhigley, line 39: This sets the values of `${self:custom.offline}` to one or the other of the sets of values defined in the preceding 6 lines

serverless.yml
```diff
@@ -42,9 +56,5 @@ package:
...
 
 functions:
-  generateRecommendations:
-    handler: src.poprox_recommender.handler.generate_recs
-    events:
-      - httpApi:
-          path: /
-          method: post
+  # if offline mode enabled filePrefix = 'offline.' else filePrefix = ''
+  ${file(./serverless/${self:custom.offline.filePrefix}functions.yaml):functions}
```
karlhigley, line 60: And then this selects between the contents of two different files depending on whether we're defining the functions for offline or deployment environments

serverless/functions.yaml
```diff
@@ -0,0 +1,9 @@
+functions:
+  generateRecommendations:
+    image:
+      name: poprox-recommender
```
karlhigley, line 4: For deployment, we use a Docker image

serverless/offline.functions.yaml
```diff
@@ -0,0 +1,8 @@
+functions:
+  generateRecommendations:
+    handler: src.poprox_recommender.handler.generate_recs
```
karlhigley, line 3: Locally/offline we just run the handler function and skip Docker

serverless.yml
```diff
@@ -1,3 +1,5 @@
+useDotenv: true
```
karlhigley, line 1: This pulls in environment variables from the env files, choosing which env file to load depending on which environment we're in (`dev` vs `prod`)

deploy.sh
```diff
@@ -0,0 +1,22 @@
...
+    esac
+done
+env=${env:-prod}
+region=${region:-us-east-1}
+
+echo "ENV: $env";
+echo "Region: $region";
+
+# Download model artifacts
+dvc pull
```
karlhigley, line 19: Mostly the same script as the other repo, but now we can use DVC to pull model artifacts

PR #10, opened 2024-05-17
Add classes for `Entity` and `Mention` domain concepts

karlhigley:


src/poprox_recommender/domain/article.py
```diff
@@ -1,12 +1,31 @@
...
+class Entity(BaseModel):
+    entity_id: Optional[UUID] = None
+    external_id: Optional[str] = None
+    name: str
+    entity_type: str
+    source: str
+    raw_data: str
+
+
+class Mention(BaseModel):
```
kluver, line 17: I'm curious why `article_id` and `mention_id` are Optional here.

src/poprox_recommender/domain/article.py
```diff
@@ -1,12 +1,31 @@
 from datetime import datetime, timezone
+from typing import List, Optional
 from uuid import UUID
 
 from pydantic import BaseModel
 
 
+class Entity(BaseModel):
+    entity_id: Optional[UUID] = None
```
kluver, line 9: Curious why `entity_id` is Optional

src/poprox_recommender/domain/article.py
karlhigley: These can exist in memory before they get inserted into the database and at that point we don't know the ids yet, because they're auto-assigned by the database on insert

src/poprox_recommender/domain/article.py
karlhigley: We could generate the UUIDs ourselves in Python if there's a reason to prefer that, but I'm most confident that Postgres will generate Postgres-compatible UUIDs

src/poprox_recommender/domain/article.py
kluver: Gotcha. That makes sense

kluver (approved):

mdekstrand (approved):

karlhigley:
I'm going to move this over to the `poprox-concepts` repo, which I hadn't planned on creating yet but turned out to be necessary in order to avoid creating a dependency on `torch` in the platform code (which then messes with the deploys)

PR #11, opened 2024-05-17
Update tokenization logic

sophiasun0515:
Finished new bert tokenizer model integration. Uploaded new model checkpoint and news-parsed test files to dvc. Tested with both offline MIND Large val dataset and serverless API endpoint and confirmed they can work.

Metrics with MIND Large val dataset: (Note the `tests/test.py` file is intended to replace the old version in PR https://github.com/CCRI-POPROX/poprox-recommender/pull/7)

> AUC: 0.6889
> MRR: 0.3324
> nDCG@5: 0.3678
> nDCG@10: 0.4333

WIP: 

- Figure out why the checkpoint performance is slightly different from original repo and our updated lolgic.
- Update test case to make sure the format (e.g. json and dict) is consistent between the serverless example and offline metric evaluation.


src/poprox_recommender/handler.py
```diff
@@ -16,29 +17,23 @@ def load_model(device_name=None):
...
+    load_path = f"{project_root()}/models/model.safetensors"
+    #with smart_open(load_path, "rb") as f:
+        #buffer = io.BytesIO(f.read())
 
-    if buffer:
-        device = th.device(device_name)
-        checkpoint = th.load(buffer, map_location=device)
+    #if buffer:
+        #device = th.device(device_name)
+        #checkpoint = th.load(buffer, map_location=device)
```
karlhigley, line 26: If we're not using these commented out lines, let's remove them—if we ever needed to recover them for some reason, we could find them in the commit history.

src/poprox_recommender/default.py
```diff
@@ -52,17 +54,27 @@ def parse_row(token_mapping, row):
...
+        return True 
+    return False
+
 # prepare the news.tsv with information for NRMS model
 def transform_article_features(articles: List[Article], token_mapping):
-    article_df = pd.DataFrame([article.model_dump() for article in articles])
+    if is_list_of_dicts: 
+        article_df = pd.DataFrame(articles)
+    else:
+        article_df = pd.DataFrame([article.model_dump() for article in articles])
```
karlhigley, line 68: I'd rather update the code in the test case to create `Article` and `ClickHistory` objects, which will make the input type supplied to this function consistent and avoid needing to check types here. Otherwise we're going to find ourselves changing `click_history.article_ids` to `click_history["article_id"]` and back again every time we want to run the test case and deploy the code.

imrecommender: I agree with you. We did not realize that the serverless examples are in JSON format. We will update the test case to use JSON format and modify the relevant parts of the code accordingly. This will ensure consistency in the input types supplied to the function and eliminate the need to switch between click_history.article_ids and click_history["article_id"] each time we run the test case and deploy the code.

karlhigley:
@sophiasun0515 @imrecommender Resolved the conflicts, looks like it might be ready to merge?

sophiasun0515:
> @sophiasun0515 @imrecommender Resolved the conflicts, looks like it might be ready to merge?

Yep looks good to go!

PR #12, opened 2024-05-24
Update MIND test data and Modify the MMR_diversify 

imrecommender:
Changes included in this PR:
1. Added `test/test_data.py` to convert MIND test data into the json format utilizing UUIDs, Article and ClickHistory classes.
2. Created `test_json.py` to evaluate the new MIND test data format using the model with tokenizer integration.
3. Fixed the MMR_diversification function.

WIP:
1. Uploading MIND test data and the UUID-ID mapping (generating UUIDs for original user and news IDs) via DVC.

### DIFF
--- a/src/poprox_recommender/test_json.py
+++ b/src/poprox_recommender/test_json.py
@@ -1,0 +1,153 @@
+import io
+import json
+import os
+import sys
+sys.path.append('../')
+import torch as th
+from safetensors.torch import load_file
+from uuid import UUID
+import numpy as np
+import pandas as pd
+from tqdm import tqdm
+from poprox_recommender.domain import Article, ClickHistory
+from poprox_recommender.default import select_articles
+from poprox_recommender.paths import project_root
+
+
+
+def load_model(device_name=None):
+
+    if device_name is None:
+        device_name = "cuda" if th.cuda.is_available() else "cpu"
+
+    load_path = f"/home/XLL1713/POPROX/engine0/NRMS_bert/NRMS_BERT_checkpoint/model.safetensors"
+    checkpoint = load_file(load_path)
+
+    return checkpoint, device_name
+
+
+def custom_encoder(obj):
+    if isinstance(obj, UUID):
+        return str(obj)
+
+def compute_mrr(y_pred, y_true):
+    # 关心原本label为1的item在y_pred中的位置 y_pred =list[排好的item id]
+    relevant_items = {item.split('-')[0] for item in y_true if item.endswith('-1')}
+    mrr = 0
+    for rank, pred_item in enumerate(y_pred, start = 1):
+        if pred_item in relevant_items:
+            mrr += (1/rank)
+    return mrr/len(relevant_items)
+
+
+
+def compute_dcg(y_pred, y_true, k=5):
+    '''
+    y_pred: 排好的item顺序
+    y_true:[item-label]
+    '''
+    item_label = {}
+    for pair in y_true:
+        item_label[pair.split('-')[0]] = int(pair.split('-')[1])
+
+    y_true = np.array([item_label[item] for item in y_pred[:k]])
+
+    gains = 2 ** y_true - 1
+    discounts = np.log2(np.arange(len(y_true)) + 2)
+
+    return np.sum(gains / discounts)
+
+
+def compute_ndcg(y_pred, y_true, k):
+    '''
+    y_pred: 排好的item顺序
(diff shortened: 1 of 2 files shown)


src/poprox_recommender/default.py
```diff
@@ -54,20 +52,19 @@ def parse_row(token_mapping, row):
     )
 
 
-# check the input type of articles 
-def is_list_of_dicts(articles): 
+# check the input type of articles
+def is_list_of_dicts(articles):
```
karlhigley, line 56: With the changes to `transform_article_features` below, I think this function can now be removed

src/poprox_recommender/default.py
```diff
@@ -141,38 +137,41 @@ def build_user_embeddings(user_df, article_embeddings, model, device, max_clicks
...
+    return S # LIST(candidate index)
 
 
 def generate_recommendations(model, articles, article_vectors, similarity_matrix, user_embeddings, num_slots=10):
     recommendations = {}
     for user, user_vector in user_embeddings.items():
         pred = model.get_prediction(article_vectors, user_vector.squeeze())
         pred = pred.cpu().detach().numpy()
-        recs = mmr_diversification(pred, similarity_matrix, theta=0.9, topk=num_slots)
+        recs = mmr_diversification(pred, similarity_matrix, theta=1, topk=num_slots)
```
karlhigley, line 174: I wouldn't mind changing the diversification coefficient, but do want to check if this is intentional since it would turn diversification off. Is that what you're aiming for?

src/poprox_recommender/default.py
```diff
@@ -222,11 +222,12 @@ def compute_similarity_matrix(todays_article_vectors):
 def select_articles(
     todays_articles: List[Article],
     past_articles: List[Article],
-    click_histories: List[ClickHistory], 
+    click_histories: List[ClickHistory],
     model_checkpoint,
     model_device,
     token_mapping,
-    num_slots=10,
```
karlhigley, line 229: Let's keep a default here if that's okay?

src/poprox_recommender/domain/click_history.py
```diff
@@ -6,4 +6,4 @@
 
 class ClickHistory(BaseModel):
     account_id: UUID = None
-    article_ids: List[UUID]
+    article_id: List[UUID]
```
karlhigley, line 9: Could we keep this as `article_ids`? I'm trying to distinguish between single values and list values with pluralization across the various domain concept classes

src/poprox_recommender/test_json.py
```diff
@@ -0,0 +1,153 @@
...
+
+    ndcg5 = []
+    ndcg10 = []
+    mrr = []
+
+    for impression_idx in range(len(mind_data)): # one by one
+
+        request_body = mind_data[impression_idx]
+
+        todays_articles = [Article.parse_obj(item) for item in request_body["todays_articles"]]
```
karlhigley, line 118: This will totally work and I'm happy to go forward with this code as is, but just FYI you can also construct these objects directly like `Article(title=mind_data["some_title_key"], url=mind_data["the_url_key"])` which would probably allow you to skip the step of converting the MIND data to JSON first

src/poprox_recommender/default.py
```diff
@@ -141,38 +137,41 @@ def build_user_embeddings(user_df, article_embeddings, model, device, max_clicks
...
     # R is all candidates (not selected yet)
-    S = []  # final recommendation (topk index)
 
+    S = []  # final recommendation (topk index)
     # first recommended item
     S.append(rewards.argmax())
 
-    for k in range(topk):
-        candidate = None
+    for k in range(topk - 1):
```
karlhigley, line 145: 👍🏻

src/poprox_recommender/default.py
```diff
@@ -141,38 +137,41 @@ def build_user_embeddings(user_df, article_embeddings, model, device, max_clicks
...
             mr_i = theta * reward_i - (1 - theta) * max_sim
-            if mr_i >= best_MR and i not in S:
+            if mr_i > best_MR:
                 best_MR = mr_i
                 candidate = i
-        if candidate:
-            S.append(candidate)
 
-    return S
+        if candidate != None:
```
karlhigley, line 164:
```suggestion
        if candidate is not None:
```

karlhigley (commented):
Looks good! I left a few questions, comments, and suggestions 🤓

PR #14, opened 2024-06-04
Deploy the recommender as a Lambda-based endpoint

karlhigley:
This sets up the relevant configuration to deploy the recommender as an independent endpoint that the platform can call instead of directly importing code from this repo. Along the way, it also removes `requirements.txt`, fixes some dependency issues, properly validates API inputs, and slims the Docker image down to a tenth of its original size which also speeds up deploys.


deploy.sh
kluver, line 1: What's going on with this file?

serverless.yml
```diff
@@ -9,7 +9,10 @@ provider:
   versionFunctions: false
   region: ${opt:region, "us-east-1"}
   stage: ${opt:stage, "local"}
-  memorySize: 512
+  memorySize: 2048
```
kluver, line 12:
```suggestion
  memorySize: 2048 # 2gb
```
On the one hand, it feels silly to document this, on the other hand, the unit isn't listed, and it's easy to think the unit might be kb, byte, or even bits.
Same with next line --- I just was not sure how to annotate that.

src/poprox_recommender/handler.py
```diff
@@ -39,8 +47,17 @@ def generate_recs(event, context):
         TOKEN_MAPPING,
         num_recs,
     )
+    for _, articles in recommendations.items():
+        for article in articles:
+            print(article)
```
kluver, line 52: debugging code, or intentional logging?

src/poprox_recommender/handler.py
```diff
@@ -39,8 +47,17 @@ def generate_recs(event, context):
...
+            print(article.model_dump())
     body = {
-        "recommendations": recommendations,
+        "recommendations": {
+            str(account_id): [
+                json.loads(article.model_dump_json()) for article in articles
```
kluver, line 57: I'm behind the ball on how this code works, but this certainly _feels_ inefficient on spec.

kluver: I'm _guessing_ that this because our models are not directly serializable and do not have a "to dictionary" type feature? If so, this is probably useful ticket for the concepts project, but an out-of-scope fix here.

kluver (approved):
Nothing I would bock on. I have a few suggested changes and a few "why is it like this?" things.

deploy.sh
karlhigley: The permissions changed to make it executable 😅

src/poprox_recommender/handler.py
karlhigley: (In essence this is a workaround for using UUIDs instead of string-ified UUIDs in the domain concept classes. I still think that's the right thing to do because it gives us validation of UUIDs, but it introduces this weird wrinkle in the serialization. Everything _almost_ works out of the box but not quite.)

karlhigley: Now that I think about it there might be a missing piece here, which would be a Pydantic model that represents the whole response instead of piecing it together from a dictionary. In that case, we could directly dump the model to JSON and bypass this whole issue. I am going to leave that for another PR, but that seems like a good pattern to adopt.

deploy.sh
kluver: oh _weird_ -- in the "changes" view it shows "empty file" but when I click through to "view file" it's not empty.

src/poprox_recommender/handler.py
kluver: I agree with the above, here's my summary from this:

- We do want to view this as a problem
- This is a "different PR" problem
- We should have a pydantic model for responses (I like this because as I think about images and our "dreams" -- we will be changing this format -- the pydantic model should help structure these changes)
- What I was imagining as a quick-n-dirty fix was just to "hide" this mess in, like a "to_dictionary" function in the base model. I _think_ having a true pydantic model for the response format resolves this issue.

PR #16, opened 2024-06-05
pfar diversification draft

imrecommender:
Update: 
- Updated the implementation of PFAR diversification. User topic preferences are captured from ClickHistory.  
- Set a parameter 'diversify' in algo_params to switch between MMR and PFAR.

TBD: 
- Test the correctness of PFAR code 
- Need to connect the topic mapping and PFAR logic together.


tests/pfar_test.py
karlhigley, line 1: If you name test files so they start with `test` (like `test_pfar.py`), then `pytest` will detect them as tests and run them as part of the test suite

karlhigley: (You can run the full test suite just by running `pytest`, or a particular test by supplying the path of the test file like this: `pytest tests/test_pfar.py`)

tests/pfar_test.py
```diff
@@ -0,0 +1,21 @@
...
+            Article.model_validate(attrs) for attrs in req_body["todays_articles"]
+        ]
+
+    topic_matched_dict, todays_article_matched_topics = match_news_topics_to_general(todays_articles)
+    print(topic_matched_dict)
+    for article_topic in todays_article_matched_topics:
+        print(todays_article_matched_topics[article_topic])
+        break
+
+if __name__ == '__main__':
```
karlhigley, line 19: And this can become a test function like:
```python
def test_pfar():
    event_path = ...
```

src/poprox_recommender/handler.py
```diff
@@ -60,3 +66,49 @@ def generate_recs(event, context):
...
+        "Lifestyle",
+        "Religion",
+        "Climate",
+        "Education",
+        "Oddities",
+    ]
+    # Load the pre-trained tokenizer and model
+    tokenizer = BertTokenizer.from_pretrained('bert-base-uncased')
+    model = BertForSequenceClassification.from_pretrained('bert-base-uncased', num_labels=len(general_topics))
+    topic_matched_dict = {} # we might be able to connect this to DB to read previous matching?
```
karlhigley, line 102: We could store matched topics in the database, but if we're going to do that the code that talks to the database to retrieve and store the matches should live in `poprox-platform`. Then the mapping would get passed in as part of the request

src/poprox_recommender/handler.py
```diff
@@ -60,3 +66,49 @@ def generate_recs(event, context):
     response = {"statusCode": 200, "body": json.dumps(body)}
 
     return response
+
+def classify_news_topic(model, tokenizer, general_topics, topic):
```
karlhigley, line 70: Something I'm confused about reading this code is how the model knows which label corresponds to which general topic. I see that the number of labels to classify with matches the number of general topics but I don't see where the model gets any general topic information it can use to figure out which labels are relevant.

src/poprox_recommender/handler.py
```diff
@@ -30,6 +33,8 @@ def generate_recs(event, context):
     todays_articles = [
         Article.model_validate(attrs) for attrs in req_body["todays_articles"]
     ]
+    topic_matched_dict, todays_article_matched_topics = match_news_topics_to_general(todays_articles)
```
karlhigley, line 36: Based on our previous finding that the general topics have 100% coverage of the stories we've seen so far, I wonder if we could ignore the rest of the topics without having to match them up. Once we add the general topics that we think are relevant to the set of topics that are already on the article, it seems like we're going to end up with the same general topics applied twice sometimes. Am I misunderstanding something about how this code works?

src/poprox_recommender/default.py
```diff
@@ -173,6 +173,53 @@ def mmr_diversification(rewards, similarity_matrix, theta, topk):
...
+            summation = 0
+
+            candidate_topic = [mention.entity.name for mention in articles[int(i)].mentions]
+            for topic in candidate_topic:
+                if topic in S_topic:
+                    product = 0
+                    break
+
+            for topic in candidate_topic:
+                summation += user_preference[topic]
```
karlhigley, line 206: Don't see where this is defined, but maybe it should be passed into this function as an argument?

src/poprox_recommender/default.py
```diff
@@ -173,6 +173,53 @@ def mmr_diversification(rewards, similarity_matrix, theta, topk):
             S.append(candidate)
     return S  # LIST(candidate index)
 
+def pfar_diversification(rewards, articles, lamb, tau, topk):
+    # articles: LIST[ARTICLE]
+    # p(v|u) + lamb*tau \sum_{d \in D} P(d|u)I{v \in d} \prod_{i \in S} I{i \in d}
+
+    S = [] # final recommendation LIST[candidate index]
```
karlhigley, line 180: Some of the information in the comment could be captured in a type annotation:
```suggestion
    S: List[int] = [] # final recommendation LIST[candidate index]
```

src/poprox_recommender/default.py
```diff
@@ -173,6 +173,53 @@ def mmr_diversification(rewards, similarity_matrix, theta, topk):
             S.append(candidate)
     return S  # LIST(candidate index)
 
+def pfar_diversification(rewards, articles, lamb, tau, topk):
```
karlhigley, line 176: To write an experiment manifest that uses both kinds of diversification we have now, we'll need to be able to provide a different URL for each one. We'll either want to set up two separate endpoints (one for MMR and one for PFAR) or a parameter on the request that switches between them.

src/poprox_recommender/handler.py
sophiasun0515: Yep that sounds good, shall we directly contribute to the `poprox-platform` repo to set up the logic? We'd appreciate if you can point us to the location to modify such info 😄

karlhigley: That is something we may have to think about a bit, because (a) I still don't totally understand why we need matching and (b) if we do need it, then there are several places it could go. One possibility would be in the code that [extracts articles from files and loads them into the database](https://github.com/CCRI-POPROX/poprox-platform/blob/main/poprox-serverless/poprox_serverless/ingest/articles.py).

src/poprox_recommender/handler.py
sophiasun0515: Ahh I see your point! Yep after reading it I do feel the general_topics is missing from the classificatino step, I just updated the logic along with a pre-fine-tuned language model based on news categories and now the test result looks much better. I'll commit the changes later.

src/poprox_recommender/handler.py
sophiasun0515: Ah if that's indeed the case, maybe we can just first check if one of the general topics is already contained in the list of topics a news have; if so, we just use it; otherwise we can trigger the classification logic?

src/poprox_recommender/handler.py
sophiasun0515: Got it! I guess I can save the matching logic now aside while first try to extract the general topic out from all available news topic lists for Xinyi to incorporate into her changes!

sophiasun0515: @imrecommender the new general news topic extraction function is `extract_general_topic()` in `poprox_recommender/hander.py`

src/poprox_recommender/handler.py
```diff
@@ -1,62 +1,110 @@
...
+def classify_news_topic(model, tokenizer, general_topics, topic):
+    inputs = tokenizer.batch_encode_plus([topic] + general_topics, return_tensors='pt', pad_to_max_length=True)
+    input_ids = inputs['input_ids']
+    attention_mask = inputs['attention_mask']
+    output = model(input_ids, attention_mask=attention_mask)[0]
+    
+    sentence_rep = output[:1].mean(dim=1)
+    label_reps = output[1:].mean(dim=1)
+
+    # now find the labels with the highest cosine similarities to the sentence
```
karlhigley, line 84: 💡 This is encoding the labels alongside the topic to be classified and then looking for similarity? I think that's what I was missing before

src/poprox_recommender/default.py
```diff
@@ -245,6 +309,44 @@ def compute_similarity_matrix(todays_article_vectors):
...
+    past_articles: List[Article],
+    click_histories: List[ClickHistory]
+):
+    '''Topic preference only based on click history'''
+    user_preference_dict = {}
+    for click_history in click_histories:
+        account_id = click_history.account_id
+        clicked_articles = click_history.article_ids # List[UUID]
+
+        topic_count_dict = {}
```
karlhigley, line 336: If you make this a `defaultdict` you can save yourself the step of checking whether keys are present or not, because all keys will get the supplied default value (zero, in this case) when they're accessed for the first time:
```
from collections import defaultdict

topic_count_dict = defaultdict(lambda: 0)
...
for topic in clicked_topics:
    topic_count_dict[topic] += 1

File: src/poprox_recommender/default.py [hunk 10]
```diff
@@ -245,6 +309,44 @@ def compute_similarity_matrix(todays_article_vectors):
...
+    for article in past_articles:
+        if article.article_id == article_id:
+            return [mention.entity.name for mention in article.mentions]
+
+def normalized_topic_count(topic_count_dict):
+
+    normalized_count = {}
+    total_count = sum(topic_count_dict.values())
+    for topic in topic_count_dict:
+        normalized_count[topic] = topic_count_dict[topic] / total_count
```
karlhigley, line 323: This will work fine as written, but you can save yourself some typing by doing it as a one-liner:
```
normalized_counts = {key: value/total_count for key,value in topic_count_dict.items()}
```

src/poprox_recommender/default.py
```diff
@@ -245,6 +309,44 @@ def compute_similarity_matrix(todays_article_vectors):
                 similarity_matrix[i, j] = similarity_matrix[j, i] = np.dot(value1, value2)
     return similarity_matrix
 
+def find_topic(past_articles: List[Article], article_id: UUID):
+    # each article might correspond to multiple topic
+    for article in past_articles:
+        if article.article_id == article_id:
+            return [mention.entity.name for mention in article.mentions]
+
+def normalized_topic_count(topic_count_dict):
```
karlhigley, line 318: I generally avoid putting types in variable names, so I'd write this with a type annotation to make it clearer what's in the dictionary:

```suggestion
def normalized_topic_count(topic_counts: dict[str, int]):
```

src/poprox_recommender/handler.py
```diff
@@ -45,3 +63,48 @@ def generate_recs(event, context):
...
     logger.info("Finished.")
     return response
+
+def extract_general_topic(article):
+    news_topics = [mention.entity.name for mention in article.mentions]
+    article_topics = set()
+    for topic in news_topics:
+        if topic in general_topics:
+            article_topics.add(topic)
+    return article_topics
```
karlhigley, line 73: This looks like it's doing set intersection on lists—correctly—but there's an easier way:
```python
general_topics = set("US News", ...)

def extract_general_topics(article: Article):
    article_topics = set([mention.entity.name for mention in article.mentions])
    return article_topics.intersection(general_topics)
```

src/poprox_recommender/handler.py
```diff
@@ -1,5 +1,7 @@
 import base64
 import logging
+from transformers import AutoTokenizer, AutoModel
```
karlhigley, line 3: I left a few comments on the changes in `handler.py`, which generally look good. I'd like to keep `handler.py` limited to the code that's processing the request and response though, so I'd suggest moving these to a new file. Maybe `topics.py`?

karlhigley:
@sophiasun0515 @imrecommender I merged the `main` branch and made some formatting adjustments just to make it easier to see your changes in this PR. Left a few comments on easier ways to do things, but this is looking good!

src/poprox_recommender/handler.py
sophiasun0515: Yep! And I manually tested on some news topics and confirmed they look accurate for sub-topic classification.

src/poprox_recommender/default.py
sophiasun0515: Good point, updated!

src/poprox_recommender/handler.py
sophiasun0515: Wow the simplification of code! TIL

src/poprox_recommender/default.py
```diff
@@ -245,6 +308,38 @@ def compute_similarity_matrix(todays_article_vectors):
...
+    past_articles: list[Article],
+    click_histories: list[ClickHistory]
+):
+    '''Topic preference only based on click history'''
+    user_preference_dict = {}
+    for click_history in click_histories:
+        account_id = click_history.account_id
+        clicked_articles = click_history.article_ids # List[UUID]
+
+        topic_count_dict = defaultdict(int)
```
karlhigley, line 332: 👍🏻

karlhigley (approved):

PR #17, opened 2024-06-07
Move model and tokenizer loading into initialization for performance

karlhigley:
Since Lambdas stay "warm" for a while and hold things in memory after a request, this makes subsequent calls to generate more recs much faster because they don't have to repeat the work of loading the model and tokenizer.

This also removes model-related code from the handler, which should make it generic across different recommenders, and consolidates it into the recommender code, which limits the number of files experimenters will have to edit.

### DIFF
--- a/src/poprox_recommender/default.py
+++ b/src/poprox_recommender/default.py
@@ -12,11 +12,12 @@ import sys
 
 sys.path.append("../")
 from tqdm import tqdm
-from nltk.tokenize import word_tokenize
+from safetensors.torch import load_file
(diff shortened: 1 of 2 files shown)


kluver (approved):
I'm not sure that every function that ended up getting moved _needs_ to be moved (I think the only _real_ change to accomplish the increased performance is moving the load model out of the function)

That said -- it looks like we got some division of "what goes where" improved along the way and I like that.

mdekstrand (approved):

PR #19, opened 2024-06-11
Rework deploy to use Conda and robust model file logic

mdekstrand:
This PR updates our image deployment setup to build from the Conda lockfile using `micromamba`, so both dev and production use the same locked dependency tree, and makes some additional reliability and quality-of-life improvements on loading model data.  Summary:

- rework `project_root()` to have error handling and work anywhere in the git repo, but fail elsewhere.
- add `model_file_path()` to locate a model file based on well-known and project root locations, to ensure it works across deployment and local installs.
- move models back to `models` (instead of `src/models`) to parallel `data`, since we now have robust code for finding them.
- commit a copy of distilbert-base-uncased to DVC and bake it into the Docker images, so it's version-locked and we aren't silently re-downloading from HuggingFace all the time. I'm not 100% happy with this solution (a git submodule might actually be better) but it gets it working and we can iterate.
- rework the Docker image to use `miniconda` to install dependencies from the conda lockfile, so we have the same frozen dependency versions across dev and deployment.
- bake the model files into a known location in the image instead of installing them in PYTHONPATH.
- build dev support through the runtime emulator into the docker image; this is auto-activated if the image is run locally.

I have tested this with `docker`. It seems to almost work with `serverless offline start`, but serverless complains that `body` is supposed to be a string. I haven't changed this because it works in Docker, but am happy to make that tweak too.


karlhigley:
You can run the Docker image locally and send it requests: https://stackoverflow.com/a/68091010

We have a [super-simple JSON file](https://github.com/CCRI-POPROX/poprox-recommender/blob/main/tests/basic-request.json) that works for quick smoke testing.

mdekstrand:
@karlhigley I think this mostly works :)

src/poprox_recommender/model/nrms/news_encoder.py
```diff
@@ -14,11 +15,11 @@ def __init__(self, config):
         self.config = config
 
         self.plm = AutoModel.from_pretrained(
-            self.config.pretrained_model, cache_dir="/tmp/"
```
karlhigley, line 17: If I understand correctly this PR bakes most (all?) of this stuff into the container, but I'd still like to keep the `cache_dir='/tmp'` as an insurance policy because the issue that solves won't bite us until we try to run in an actual deployed environment

karlhigley (approved):

src/poprox_recommender/model/nrms/news_encoder.py
mdekstrand: Yep, the PR bakes this all into the container, but I'm happy to leave the cache dir there.

mdekstrand: @karlhigley Pushed a change that re-adds it, but using `tempfile.gettempdir()` instead of hard-coding `/tmp` — sound good?

PR #20, opened 2024-06-12
Allow passing algorithm params in as query string parameters

karlhigley:
This provides a generic way to supply algorithm parameters as part of the request URL across a variety of algorithms. The code in `handler.py` knows nothing about the actual recommender, but still hands off an algorithm specific parameter (theta) that tunes the amount of diversification. (The code in `default.py` is the algorithm-specific part that knows which parameters to look for.)

### DIFF
--- a/src/poprox_recommender/handler.py
+++ b/src/poprox_recommender/handler.py
@@ -1,4 +1,5 @@
 import json
+import logging
 
 from poprox_concepts.api.recommendations import (
     RecommendationRequest,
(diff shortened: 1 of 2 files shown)


src/poprox_recommender/handler.py
```diff
@@ -1,20 +1,32 @@
...
+import logging
 
 from poprox_concepts.api.recommendations import (
     RecommendationRequest,
     RecommendationResponse,
 )
 from poprox_recommender.default import select_articles
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
```
mdekstrand, line 11: I'd rather see this in a different logging configuration place, but cleaning up log config can be a separate PR (which I was thinking of doing anyway, because I was looking for logging while debugging #19).

mdekstrand (approved):

src/poprox_recommender/handler.py
karlhigley: Yeah, no strong opinions about how logging ought to be set up. Mostly haven't worried about it at all thus far but needed a way to make it visible which params were getting received in order to know this was working

PR #30, opened 2024-06-14
Revise path logic and move models and data to top-level

mdekstrand:
This updates the path-location logic and rearranges the repository structure to be more logical and reduce wasted space during Docker image build:

- establishes top-level `src`, `data`, `models` directories for source code, input data, and model checkpoints, respectively.
- adds `model_file_path` function to locate a model file, checking in a `POPROX_MODELS` environment variable, the project root, and the Conda environment prefix, in that order. This will enable POPROX model data to be stored in a data location during deployment.
- restores `project_root` and makes it search based on the current working directory, looking for `pyproject.toml`, so that it can work seamlessly regardless of whether and how `poprox_recommenders` has been installed.

It also adds tests for the `paths` module, and depends on #31 and the docker tests to make sure none of this breaks our code in a way that can be detected at test time.


src/poprox_recommender/paths.py
```diff
@@ -1,5 +1,107 @@
...
+            _logger.info("resolved %s: %s", name, mf)
+            return mf
+
+    msg = f"model file {name}"
+    raise FileNotFoundError(msg)
 
 
-def src_dir() -> Path:
-    return Path(__file__).parent.parent
+def _is_project_root(path: Path) -> bool:
```
karlhigley, line 90: If finding the project root is going to involve parsing a file (which seems like a lot of work), it should probably be stored somewhere so we don't have to parse the file more than once.

src/poprox_recommender/test_offline.py
```diff
@@ -36,7 +36,7 @@ def recsys_metric(recommendations, row_index, news_struuid_ID):
     # use the url of Article
     impressions_truth = (
         pd.read_table(
-            src_dir() / "data" / "test_mind_large" / "behaviors.tsv",
+            f"{project_root()}/data/test_mind_large/behaviors.tsv",
```
karlhigley, line 39: Let's keep the pathlib style here

karlhigley (approved):

PR #32, opened 2024-06-14
Use `pathlib` for paths instead of strings for Windows compatibility

karlhigley:
Although this still looks like it uses forward slashes, that's just `pathlib`'s way of saying "path separator," and it will translate them to the appropriate backslashes on Windows.

### DIFF
--- a/src/poprox_recommender/test_offline.py
+++ b/src/poprox_recommender/test_offline.py
@@ -20,7 +20,7 @@ def load_model(device_name=None):
     if device_name is None:
         device_name = "cuda" if th.cuda.is_available() else "cpu"
 
-    load_path = f"{src_dir()}/models/model.safetensors"
+    load_path = src_dir() / "models" / "model.safetensors"
     checkpoint = load_file(load_path)
 
     return checkpoint, device_name
@@ -36,7 +36,7 @@ def recsys_metric(recommendations, row_index, news_struuid_ID):
     # use the url of Article
     impressions_truth = (
         pd.read_table(
-            f"{src_dir()}/data/test_mind_large/behaviors.tsv",
+            src_dir() / "data" / "test_mind_large" / "behaviors.tsv",
             header="infer",
             usecols=range(5),
             names=["impression_id", "user", "time", "clicked_news", "impressions"],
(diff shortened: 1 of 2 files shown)


src/poprox_recommender/test_offline.py
```diff
@@ -69,11 +69,11 @@ def recsys_metric(recommendations, row_index, news_struuid_ID):
     MODEL, DEVICE = load_model()
     TOKEN_MAPPING = "distilbert-base-uncased"  # can be modified
 
-    with open(f"{src_dir()}/data/val_mind_large/news_uuid_ID.json") as json_file:
+    with open(str(src_dir() / "data" / "val_mind_large" / " news_uuid_ID.json")) as json_file:
```
mdekstrand, line 72: Don't need `str` — `open` can take a `Path`. (and when you do need strings because some API doesn't know about paths, `os.fspath` is the robust, cross-platform, cross-nonsense way to turn a `Path` into an appropriate string.)

src/poprox_recommender/test_offline.py
```diff
@@ -69,11 +69,11 @@ def recsys_metric(recommendations, row_index, news_struuid_ID):
...
         news_struuid_ID = json.load(json_file)
 
     # load the mind test json file
-    with open(f"{src_dir()}/data/val_mind_large/mind_test.json") as json_file:
+    with open(str(src_dir() / "data" / "val_mind_large" / "mind_test.json")) as json_file:
```
mdekstrand, line 76: same remark as above.

src/poprox_recommender/test_offline.py
```diff
@@ -20,8 +20,8 @@ def load_model(device_name=None):
     if device_name is None:
         device_name = "cuda" if th.cuda.is_available() else "cpu"
 
-    load_path = f"{src_dir()}/models/model.safetensors"
-    checkpoint = load_file(load_path)
+    load_path = src_dir() / "models" / "model.safetensors"
+    checkpoint = load_file(str(load_path))
```
mdekstrand, line 24: same remark as below

src/poprox_recommender/test_offline.py
```diff
@@ -36,7 +36,7 @@ def recsys_metric(recommendations, row_index, news_struuid_ID):
     # use the url of Article
     impressions_truth = (
         pd.read_table(
-            f"{src_dir()}/data/test_mind_large/behaviors.tsv",
+            str(src_dir() / "data" / "test_mind_large" / "behaviors.tsv"),
```
mdekstrand, line 39: same remark as below

mdekstrand (approved):
Drop `str` and LGTM.

src/poprox_recommender/test_offline.py
karlhigley: For feedback like this, I highly recommend the suggestion button 😄 
Screenshot-20240614174233-67x53

PR #35, opened 2024-06-17
Integrate onboarding topic selections into PFAR diversification

karlhigley:
Companion PRs:
* https://github.com/CCRI-POPROX/poprox-concepts/pull/6
* https://github.com/CCRI-POPROX/poprox-platform/pull/78

This combines onboarding selections and the topics of clicked articles into a single topic distribution. The five-point scale presented during onboarding is adjusted so that the lowest interest level counts as zero clicks and the highest counts as four clicks, which are then summed with the topics from clicked articles and then normalized before being passed into the PFAR implementation.

PR #36, opened 2024-06-19
Add script to automate reliable relocking

mdekstrand:
This adds a shell script that runs `conda-lock` with the correct inputs and environment variables to ensure that locking is successful regardless of the user's local Conda configuration. It also tweaks a couple of dependency specifications to make sure that the `main` category contains a closure (`conda-lock` is not 100% on achieving that).

Once the upstream conda-lock bug is fixed, we can update this script to add the ability to update only our dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,19 @@ To get the data and models, there are two steps:
 1.  Obtain the credentials for the S3 bucket and put them in `.env` (the environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`)
 2.  `dvc pull`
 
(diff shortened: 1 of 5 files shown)


mdekstrand:
This is failing the Docker test, because it does not have the code updates to use the `poprox-concepts` updates (those are in #35 I think), and the Docker image build does not use locked dependencies.

Once we use lockfiles for Docker image dependencies (which in the current setup would cause both Docker and local tests to fail in #35, not just local), then we'll be in a position where deployment and tests both reliably use the locked dependencies, and are consistent with local development.

karlhigley (approved):

PR #37, opened 2024-06-19
Use Conda to build Docker image from locked dependencies

mdekstrand:
This PR updates our Docker build to use Conda to build the image from the locked dependencies.

This solves the problem that our main builds are not using locked dependencies for deployment, so it is possible for deployment of `main` to succeed one day but fail later because of external dependency updates. By using locked dependencies, it ensures (mostly) reproducible and durable deployability of the main branch.


karlhigley:
I want to deploy the state of `main` before this change to make sure that's working before merging this and rolling it out

karlhigley (approved):

PR #38, opened 2024-06-24
Fix a set insertion bug in PFAR implementation

karlhigley:
As originally written, this was inserting generator expressions into the set and comparing them against topic names which never matched. That resulted in PFAR choosing articles that were all on the same topic, because it never found the candidate topics in the set of seen topics.


sophiasun0515 (approved):
lgtm!

PR #39, opened 2024-06-24
Compute PFAR's per-user tau parameter if it isn't supplied

karlhigley:
If `pfar_tau` isn't included in the query parameters, we can compute it with information entropy according to the paper


sophiasun0515 (approved):
Looks good! Thanks for catching this edge case

PR #40, opened 2024-06-24
Use the high-level general topics that match onboarding to diversify

karlhigley:
The long list of AP topics has a bunch of very fine-grained stuff in it that makes articles look more topically diverse than they really are, but we know that our short list of high-level topics has close to 100% coverage of the stories we've seen so we can use it here to diversify over a narrower set.


sophiasun0515 (approved):
looks good!

PR #41, opened 2024-06-24
Pass relevance scores through a sigmoid before PFAR

karlhigley:
PFAR is expecting probabilities in the range 0 to 1, and the tuning parameters don't quite work as expected otherwise. One option is to squash our dot products into that range with a sigmoid and interpret the result as a probability that a user is interested in an item


sophiasun0515 (approved):
🚢

PR #43, opened 2024-06-24
Break the recommendation process apart into independent-ish components

karlhigley:
This doesn't take us all the way to a pipeline abstraction or even a standard component interface, but it helps clarify what the various components we already have in play do and how they fit together. There aren't any (intentional) changes to the functionality here; this just re-organizes the code to make it easier to think about.

### DIFF
--- a/src/poprox_recommender/model/__init__.py
+++ b/src/poprox_recommender/model/__init__.py
@@ -1,0 +1,50 @@
+from dataclasses import dataclass
+
+from safetensors.torch import load_file
+from transformers import AutoTokenizer
+
+from poprox_recommender.model.nrms import NRMS
+from poprox_recommender.paths import model_file_path
+
+
+@dataclass
+class ModelConfig:
+    num_epochs: float = 10
+
+    num_clicked_news_a_user: float = 50
+    word_freq_threshold: float = 1
+    dropout_probability: float = 0.2
+    word_embedding_dim: float = 300
+    category_embedding_dim: float = 100
+    query_vector_dim: float = 200
+    additive_attn_hidden_dim: float = 200
+    num_attention_heads: float = 16
+    hidden_size: int = 768
+
+    pretrained_model = "distilbert-base-uncased"
+
+
+def load_checkpoint(device_name=None):
+    checkpoint = None
+
+    if device_name is None:
+        # device_name = "cuda" if th.cuda.is_available() else "cpu"
+        device_name = "cpu"
+
+    load_path = model_file_path("model.safetensors")
+
+    checkpoint = load_file(load_path)
+    return checkpoint, device_name
+
+
+def load_model(checkpoint, device):
+    model = NRMS(ModelConfig()).to(device)
+    model.load_state_dict(checkpoint)
+    model.eval()
+
+    return model
+
+
+TOKENIZER = AutoTokenizer.from_pretrained("distilbert-base-uncased", cache_dir="/tmp/")
+CHECKPOINT, DEVICE = load_checkpoint()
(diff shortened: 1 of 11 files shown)


sophiasun0515 (approved):
Looks good and I like how we refactor different object classes into a more organized structure!

src/poprox_recommender/model/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+    model = NRMS(ModelConfig()).to(device)
+    model.load_state_dict(checkpoint)
+    model.eval()
+
+    return model
+
+
+TOKENIZER = AutoTokenizer.from_pretrained("distilbert-base-uncased", cache_dir="/tmp/")
+CHECKPOINT, DEVICE = load_checkpoint()
+MODEL = load_model(CHECKPOINT, DEVICE)
```
mdekstrand, line 50: One of the things I have been wanting to do, and will defer until we're ready to merge these changes, is to move model initialization out of import and into runtime (or possibly import-time in `handlers`, but not in `model`). The idea is to move all of these steps into one load-the-model function and full model object (even if it's just a named tuple of components), so that we set up the models when code is run, not when it's imported (this will e.g. speed up PyTest test collection, because right now pytest will load checkpoints and set up models before it's even figured out what tests to run).

All of that to get to a question — do we want to do that as a part of this refactor PR, or save for a follow-up PR?

mdekstrand (approved):
I really like what this is doing, and it sets things up for some additional changes I have been wanting to work on. I will defer both those changes (one of which may make sense to roll into this PR), and baking the HuggingFace models into our repo, until we have meged this.

src/poprox_recommender/model/__init__.py
karlhigley: That change would revert an earlier change that we made in order to speed up the Lambda in production. If we do it at run-time, we have to reload the model every time the Lambda is called, which is slow and leads to timeouts. If we do it at import, the model state is kept around after the first call as long as the Lambda is "warm." Once we hit a gap in the requests, then the underlying resources are de-provisioned and we have to load the model again on the next request, so we start our batch of requests each day from a "cold" state but warm up after the first one.

karlhigley: (There may be a way to thread this needle, but I haven't figured out what that is yet.)

mdekstrand: I see 3 reasonable solutions here:

1. Load at import, but only in `handler.py`, not the model definition files — `handler.py` just calls our single `load_model` or whatever function and stores the result in a global. This preserves current lambda behavior while not pre-loading the model for tests / CLI operations / interactive workflows.
2. Cache model instantiation, with a function that stashes the instantiated model in a global the first time it is called and returning the global on later calls. I use this pattern quite frequently in my own code.
3. If Lambda supports setup functions (no idea here, they might just expect import to do this, which is fine): instantiate the model in the setup function.

mdekstrand: If you want me to take a swing at one of these as a separate effort after this PR is merged that's fine with me.

karlhigley: I think the best option here is likely 2—I'm not aware of support for setup functions beyond "whatever runs on import", and the docs seems to suggest that [caching in globals should work](https://docs.aws.amazon.com/lambda/latest/operatorguide/static-initialization.html).

The model loading used to be in the handler, but my theory is that the handler shouldn't know anything about the recommender other than which function to call to get recommendations, because that makes it generic across recommenders (and therefore not something experimenters should need to worry about editing.)

PR #44, opened 2024-06-25
Simplify article embedding code

karlhigley:
This removes `pandas` and `swifter` as (non-dev) dependencies, and may also fix a bug in the existing embedding code (judging by the recommendations produced.)


mdekstrand (approved):
glad to see the dep dropped. and that's not a small drop in dependency weight.

sophiasun0515 (approved):
:shipit:

PR #45, opened 2024-06-25
Use equal weights for unselected topics in PFAR

karlhigley:
If I understand correctly, the original paper assumes that people prefer providers equally, which we know isn't true when the "providers" are topics. However, limiting the PFAR boost to topics people have actually clicked on (or selected during onboarding) reduces the impact of diversification to only one or two slots out of the ten in our newsletters.

I think the reason is that the set of selected topics grows very quickly since most articles have several high-level topics, so after the first few slots there simply aren't any more topics that haven't already been selected. That's even more true when using the intersection of people's topic preferences with the list of high-level topics (which will be a shorter list.)

### DIFF
--- a/src/poprox_recommender/default.py
+++ b/src/poprox_recommender/default.py
@@ -17,7 +17,7 @@ from safetensors.torch import load_file
 from tqdm import tqdm
 from transformers import AutoTokenizer
 
-from poprox_concepts import Article, ClickHistory, InterestProfile
+from poprox_concepts import GENERAL_TOPICS, Article, ClickHistory, InterestProfile
 from poprox_recommender.model.nrms import NRMS
 from poprox_recommender.paths import model_file_path
 from poprox_recommender.topics import extract_general_topics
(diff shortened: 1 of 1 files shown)


rburke2233:
There is a version of PFAR that wasn't published in that paper but that we used in subsequent experimentation where the indicator function is replaced by a metric in the range [0..1]. Think of it as the desired concentration of items of that category. In the current formalization, as soon as a single "protected" item appears in the list, the boost for those items goes to zero. And, as you say, if you only have a couple of protected categories, then the diversity term ends up always at zero. This is an easy change technically; the harder question is deciding what the metric should look like. In our prior work, we set a target percentage and had the metric measure (one-sided) distance from that target.

karlhigley:
One potential adjustment I played with a bit was to apply the boost for each of an item's topics that aren't already in the selected topic set instead of zeroing the boost if there's any overlap between the selected topic set and the item's topics. That seemed to produce recommendations with less outliers in topics I'm not interested in but also less diverse recommendations on the whole.

rburke2233:
Effectively you're making this an "AND" rather than an "OR", which makes perfect sense. PFAR wasn't designed with the idea of that items would be associated with a vector of categories across which diversification would take place. It was designed for fairness situations where there would be a binary category: protected / unprotected per item.

karlhigley:
Closing this since the PR has been overtaken by other changes, but I'll open a new one and link it here

PR #47, opened 2024-06-27
Update "no clicks" default to match onboarding topic selections

karlhigley:
When users sign up, their first newsletter (and any subsequent newsletters before they click an article) should reflect the topics they chose during onboarding. Here we filter the available articles to just those topics and randomly sample from the remaining articles as a simple stop-gap.


karlhigley:
@sophiasun0515 Yeah, totally open to changing this to account for degree of preference somehow, but I couldn't think of a simple way to do it off the top of my head. Y'all may have some ideas though and I don't think the way I implemented here is particularly good, so feel free to improve on this!

PR #51, opened 2024-07-01
Use equal weights for unselected topics in PFAR

karlhigley:
(Based on #45 and the updated state of `main`)

If I understand correctly, the original paper assumes that people prefer providers equally, which we know isn't true when the "providers" are topics. However, limiting the PFAR boost to topics people have actually clicked on (or selected during onboarding) reduces the impact of diversification to only one or two slots out of the ten in our newsletters.

I think the reason is that the set of selected topics grows very quickly since most articles have several high-level topics, so after the first few slots there simply aren't any more topics that haven't already been selected. That's even more true when using the intersection of people's topic preferences with the list of high-level topics (which will be a shorter list.)

### DIFF
--- a/src/poprox_recommender/diversifiers/pfar.py
+++ b/src/poprox_recommender/diversifiers/pfar.py
@@ -2,7 +2,7 @@ import math
 
 import torch as th
 
-from poprox_recommender.topics import extract_general_topics, normalized_topic_count
+from poprox_recommender.topics import GENERAL_TOPICS, extract_general_topics, normalized_topic_count
 
 
 class PFARDiversifier:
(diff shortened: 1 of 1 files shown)


sophiasun0515:
I get your point. So now the implementation would treat the effect from each general topic equally for diversification purpose? It technically looks sound, and we can test internally to see if that can help with actual user satisfaction.

karlhigley:
Yeah, this is the version that's been in our emails and the one that aligns most closely with the paper. It seems less well-motivated by what we know about topic preferences, but it produces more diversified recs.

sophiasun0515 (approved):

PR #52, opened 2024-07-01
Move article embedding lookup table construction to `UserEmbedder`

karlhigley:
The lookup table is only used in the creation of user embeddings, so we can defer constructing it until we actually need it. That avoids the need to return two different types that contain the same information in different formats from the `ArticleEmbedder`, which cleans up the calling code a bit.


karlhigley:
Yes, but it goes the other way: all of these changes are in #53 but none of #53's changes are included here

PR #53, opened 2024-07-02
Standardize component argument and return types

karlhigley:
This builds on the previous refactoring and CCRI-POPROX/poprox-concepts#9 to standardize the component API a bit. After this refactor, every component takes an `ArticleSet` (an extendable wrapper around a list of articles) as the first argument and some take an `InterestProfile` (an extendable wrapper around click histories etc) as the second argument. Additionally, all components return one of these two types.

The mechanism by which components extend either `ArticleSet` or `InterestProfile` with new information that isn't part of their default definitions is Pydantic's ability to accept and set additional attributes beyond those in the class definition, which components can use to create new attributes like `embeddings`, `scores`, etc. I'd like to find a way to make these more visible, both to readers of the code and to the code itself so that we can validate that a sequence of components produces everything later components require before they try to access it. For now though, I'm leaving it very implicit in the interest of keeping the scope of this change manageable.

This doesn't take us all the way to a pipeline, but it comes pretty close—the next step along the path is likely to execute a list of components in sequence while forming the inputs for the next component from a combination of the previous component's inputs and output. Once we make it that far, then we can start to think about more complex configurations like branching component graphs etc.

### DIFF
--- a/conda-lock.yml
+++ b/conda-lock.yml
@@ -77,11 +77,11 @@ package:
   manager: conda
   platform: osx-arm64
   dependencies:
+    aiohttp: '>=3.9.2,<4.0.0'
+    aioitertools: '>=0.5.1,<1.0.0'
+    botocore: '>=1.34.70,<1.34.132'
     python: '>=3.8'
     wrapt: '>=1.10.10,<2.0.0'
-    aioitertools: '>=0.5.1,<1.0.0'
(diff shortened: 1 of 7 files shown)


mdekstrand (approved):

sophiasun0515 (approved):
🚢

PR #54, opened 2024-07-02
Dynamically cache the recommender at runtime

mdekstrand:
This moves model loading into a `get_model` function to defer loading from import to runtime, caching in a global variable. This returns both the tokenizer and scoring model.

### DIFF
--- a/src/poprox_recommender/model/__init__.py
+++ b/src/poprox_recommender/model/__init__.py
@@ -9,7 +9,7 @@ from transformers import AutoTokenizer
 from poprox_recommender.model.nrms import NRMS
 from poprox_recommender.paths import model_file_path
 
-_cached_recommender: Optional[RecommenderComponents] = None
+_cached_model: Optional[RecommenderComponents] = None
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


karlhigley:
Looks good, with one note: I wouldn't call the model/device/tokenizer a recommender, because conceptually the recommender includes things like which reranker/diversifier get applied (if any.) Maybe `get_model` etc?

src/poprox_recommender/model/__init__.py
```diff
@@ -24,6 +29,13 @@ class ModelConfig:
     pretrained_model = "distilbert-base-uncased"
 
 
+@dataclass
+class RecommenderComponents:
+    tokenizer: AutoTokenizer
+    model: NRMS
```
sophiasun0515, line 35: Curious if we need a more general class for the model object (like a parent class for both NRMS, PFAR, etc.). But otherwise looks good!

sophiasun0515 (approved):

src/poprox_recommender/model/__init__.py
mdekstrand: Eventually, yes, for now just want to capture what we're currently doing as a minimal increment.

mdekstrand: Once @karlhigley has gotten farther on the refactor + pipelining I expect this class will go away and we'll just have a function that returns the whole pipeline.

PR #55, opened 2024-07-02
Bake HuggingFace models into DVC + Docker

mdekstrand:
This adds the HuggingFace models directly to our DVC repo, bakes them into the Docker image, and loads them from there instead of directly from HF behind the scenes.

This improves the situation in a few ways:

* Our HuggingFace models are version-locked so they can't change underneath us.
* All resources are tracked directly instead of fetched through behind-the-scenes HTTP downloads.
* Startup is probably faster because it doesn't need to re-fetch a half-gig of model data from HF.


src/poprox_recommender/model/__init__.py
```diff
@@ -9,6 +9,7 @@
 from poprox_recommender.model.nrms import NRMS
 from poprox_recommender.paths import model_file_path
 
+TOKENIZER_NAME = "distilbert-base-uncased"
```
karlhigley, line 12:
```suggestion
MODEL_NAME = "distilbert-base-uncased"
```

src/poprox_recommender/model/__init__.py
```diff
@@ -26,7 +27,7 @@ class ModelConfig:
     num_attention_heads: float = 16
     hidden_size: int = 768
 
-    pretrained_model = "distilbert-base-uncased"
+    pretrained_model = TOKENIZER_NAME
```
karlhigley, line 30:
```suggestion
    pretrained_model = MODEL_NAME
```

src/poprox_recommender/model/__init__.py
```diff
@@ -60,7 +61,7 @@ def load_model(checkpoint, device):
 def get_model(device=None) -> RecommenderComponents:
     global _cached_model
     if _cached_model is None:
-        tokenizer = AutoTokenizer.from_pretrained("distilbert-base-uncased", cache_dir="/tmp/")
+        tokenizer = AutoTokenizer.from_pretrained(model_file_path(TOKENIZER_NAME), cache_dir="/tmp/")
```
karlhigley, line 64:
```suggestion
        tokenizer = AutoTokenizer.from_pretrained(model_file_path(MODEL_NAME), cache_dir="/tmp/")
```

src/poprox_recommender/topics.py
```diff
@@ -6,6 +6,9 @@
 
 from poprox_concepts import Article, ClickHistory
 from poprox_concepts.domain.topics import GENERAL_TOPICS
+from poprox_recommender.paths import model_file_path
+
+NEWS_MODEL = "news-category-classifier-distilbert"
```
karlhigley, line 11: Though the code for topic matching exists in the repo, I don't think we're using it at all right now. Baking this model into the container image is an extra 400MB we don't currently need AFAIK

karlhigley (approved):

src/poprox_recommender/topics.py
mdekstrand: This worked.

mdekstrand:
@karlhigley Just merged this, see Slack for one outstanding issue the final testing run-up surfaced.

PR #56, opened 2024-07-03
Create and apply a `RecommendationPipeline` class

karlhigley:
This introduces a simple-ish `RecommendationPipeline` that works with the components created in #43 and #53. Conceptually, the components run in the order they're added to the pipeline and modify shared state that each component can select inputs from and write outputs to. It could be generalized (e.g. to support types other than `ArticleSet` and `InterestProfile`), but this is enough to support the recommenders we're currently building/using.


tests/basic-request.json
```diff
@@ -1,28 +1,28 @@
...
+  "todays_articles": [
+    {
+      "article_id": "7e5e0f12-d563-4a60-b90a-1737839389ff",
+      "title": "title 2",
+      "content": "content 2",
+      "url": "url 2"
+    }
+  ],
+  "interest_profile": {
+    "profile_id": "9eb70e32-fff9-4e8a-a05b-df582174fbf5",
```
karlhigley, line 19: We use `profile_id` instead of `account_id` here because the id supplied by the platform side may not actually be the account id if we're running an experiment recommender since we aren't allowed to track accounts across experiments.

src/poprox_recommender/filters/topic.py
```diff
@@ -0,0 +1,20 @@
+from poprox_concepts import ArticleSet, InterestProfile
+
+
+class TopicFilter:
```
karlhigley, line 4: I haven't yet introduced tests for the various new components, but they should be much easier to write tests for now that they're modular and decoupled from the main flow of the recommendation code

src/poprox_recommender/pipeline.py
```diff
@@ -0,0 +1,56 @@
...
+        if not isinstance(recs, ArticleSet):
+            msg = f"The final pipeline component must return ArticleSet, but received {type(recs)}"
+            raise TypeError(msg)
+
+        return recs
+
+    def run_component(self, component_spec: ComponentSpec, state: dict[str, ArticleSet | InterestProfile]):
+        arguments = []
+        for input_name in component_spec.inputs:
+            arguments.append(state[input_name])
```
karlhigley, line 43: In theory we should validate the types of the inputs and raise `TypeError`s here but I suspect that might be code we'll just rip back out. I didn't hit any cases where clearer error messages on inputs would have helped me, but I could be persuaded to add them if it seems useful.

src/poprox_recommender/default.py
```diff
@@ -1,91 +1,55 @@
...
     else:
-        return random.sample(topical_articles, len(topical_articles)) + random.sample(
-            other_articles, num_slots - len(topical_articles)
-        )
+        topic_filter = TopicFilter()
+        sampler = UniformSampler(num_slots=num_slots)
+
+        pipeline = RecommendationPipeline(name="random_topical")
+        pipeline.add(topic_filter, inputs=["candidate", "profile"], output="topical")
+        pipeline.add(sampler, inputs=["topical", "candidate"], output="recs")
```
karlhigley, line 47: In the future, I'd like to make these recommendation pipelines serializable to/from YAML (or similar.) Then the handler could read all the recommender definition files present in a directory, load the pipelines into a dictionary by name, and be prepared to serve any of them based on a query string parameter. That would give us an easy way to map from the names of the recommenders in the experiment manifest to the actual pipeline that matches it, and allow people to define recommenders without writing any Python (as long as they use existing components we provide.)

tests/test_select.py
```diff
@@ -37,17 +39,31 @@ def test_select_by_topic_filters_articles():
...
+    pipeline.add(sampler, inputs=["topical", "candidate"], output="recs")
+
     # If we can, only select articles matching interests
-    recs = select_by_topic(articles, profile, num_slots=2)
+    inputs = {
+        "candidate": ArticleSet(articles=articles),
+        "clicked": ArticleSet(articles=[]),
+        "profile": profile,
+    }
+    recs = pipeline(inputs)
```
karlhigley, line 55: The code above is repeated from `default.py`, which we'd ideally avoid by either restructuring the construction of recommender pipelines to make it re-usable or by reading the pipeline configuration from a file.

karlhigley:
I'm having a hard time with the PFAR tests, which are super-slow locally and seem to be getting killed (OOM?) in the Github Action

src/poprox_recommender/default.py
```diff
@@ -1,91 +1,55 @@
...
+
+    return pipeline(
+        {
+            "candidate": candidate_articles,
+            "clicked": clicked_articles,
+            "profile": interest_profile,
+        }
+    )
```
karlhigley, line 55: The pipeline currently just returns the output of the last component, but another possibility—which may be useful for e.g. computing diversity metrics that compare the original vs reranked recs—would be to optionally specify which additional elements of the pipeline state you want to get back (if any.)

src/poprox_recommender/default.py
mdekstrand: Given the amount of machinery needed to handle initialization, checkpointing, training, etc., I would probably put this as a relatively low priority for poprox.

mdekstrand (approved):
I'm pretty happy with this for a basic non-DAG pipeline — makes sense, and gets the job done in a reasonable and simple way.

My one substantive suggestion is to rethink the use of positional parameters vs. keyword parameters for specifying component inputs. I think a keyword-based design could provide more opportunity to reduce risk of error.

src/poprox_recommender/default.py
karlhigley: I suspect we see this differently because you're thinking of a pipeline that handles all the work required to prepare models and I'm not, particularly because support for all that machinery across multiple frameworks (TF, Torch, JAX, LensKit, etc) seems fairly impossible given our constrained resources.

We're going to have a default model (or several) that are pre-trained and available to experimenters, and I'm told that many people may want to use those models in recommenders that only modify things like presentation logic only. So, with this idea I'm responding to earlier feedback that forking a repo and modifying code is going to be much harder for e.g. social scientists than hitting an endpoint. (You may remember my perspective to the contrary was described as an example of my "developer brain" not having empathy for non-developers.)

src/poprox_recommender/pipeline.py
```diff
@@ -0,0 +1,56 @@
...
+    inputs: list[str]
+    output: str
+
+
+class RecommendationPipeline:
+    def __init__(self, name):
+        self.name = name
+        self.components = []
+
+    def add(self, component: Callable, inputs: list[str], output: str):
```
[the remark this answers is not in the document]
karlhigley: Yeah, I like this. Let's try it!

src/poprox_recommender/pipeline.py
karlhigley: his was the quickest, cheapest pipeline I could come up with so my main thought was "this is a smaller, simpler step forward," but I wouldn't be opposed to moving toward keyword arguments, better type-checking, etc

src/poprox_recommender/default.py
karlhigley: "Here's a library of pre-fab semi-configurable recommender components, here's a text-editable file format that defines executable recommenders, here's how to use it to build experiment recommenders" seems like the kind of thing that might add value for our target audience (and might help with our "what exactly do we document?" conundrum too)

karlhigley:
@mdekstrand This is sort of half-way to a DAG, in the sense that the inputs and outputs can be wired up into a directed graph but the definition order and the execution order are still coupled. From a certain perspective, the execution order is kind of an implementation detail (though it does leak through the abstraction a bit) because we could change that to e.g. add concurrent execution without modifying the API. All of which is to say: I think this can grow into something that feels more like other DAG APIs, but I agree that it's not there yet.

src/poprox_recommender/pipeline.py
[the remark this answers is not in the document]
karlhigley: One wobble: going from `output` to `name` makes it less clear what's happening and how to name things. For example, the existing pipeline definition for our default recommender starts to look wrong, because it seems like multiple components are getting the same name:
```
pipeline = RecommendationPipeline(name=diversify)
pipeline.add(article_embedder, name="candidate", inputs=["candidate"])
pipeline.add(article_embedder, name="clicked", inputs=["clicked"])
pipeline.add(user_embedder, name="profile", inputs=["clicked", "profile"])
pipeline.add(article_scorer, name="candidate", inputs=["candidate", "profile"])
pipeline.add(diversifier, name="recs", inputs=["candidate", "profile"])
```

But giving the outputs different names makes it confusing whether the names refer to the components or the outputs, what the types of the outputs are, and what they included. Does `user_embed` output the whole interest profile or just the user embedding?
```
pipeline = RecommendationPipeline(name=diversify)
pipeline.add(article_embedder, name="cand_embed", inputs=["candidate"])
pipeline.add(article_embedder, name="click_embed", inputs=["clicked"])
pipeline.add(user_embedder, name="user_embed", inputs=["click_embed", "profile"])
pipeline.add(article_scorer, name="scored_articles", inputs=["cand_embed", "user_embed"])
pipeline.add(diversifier, name="recs", inputs=["scored_articles", "user_embed"])
```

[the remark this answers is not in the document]
karlhigley: Another wobble: the API now depends on how individual components name their parameters, which may not be consistent across components (especially as experimenters start writing their own.)

```
pipeline = RecommendationPipeline(name=diversify)
pipeline.add(article_embedder, output="candidate", article_set="candidate")
pipeline.add(article_embedder, output="clicked", article_set="clicked")
pipeline.add(user_embedder, output="profile", clicked_articles="clicked", interest_profile="profile")
pipeline.add(article_scorer, output="candidate", candidate_articles="candidate", interest_profile="profile")
pipeline.add(diversifier, output="recs", candidate_articles="candidate", interest_profile="profile")
```

karlhigley:
With the intention of continuing to discuss improvements and pursuing further iteration, I think I'm going to go ahead and merge on the theory that this is better than what we had before and a step in the right general direction. In particular, @rburke2233 has an interest in retrieving the outputs of intermediate steps of the recommendation process, which I think we can build on top of these changes more easily than without them (even if the pipeline/component API continues to evolve.)

src/poprox_recommender/pipeline.py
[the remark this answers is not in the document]
mdekstrand: First wobble: I'm fine with keeping `output` and making it a keyword parameter.

Second wobble: I see that as a bit of a feature, actually:

* It may encourage consistency
* It raises a failure that results in doc-consultation (mismatched name, so one looks at the docs, hopefully)
* IMO it's a worthwhile tradeoff to detect / prevent the more subtle and harder-to-detect error where two components do take the same inputs, but in a different order

src/poprox_recommender/default.py
mdekstrand: Makes sense. If we're very tightly-scoped I can see it working.

The challenge I see — setting aside training and generating checkpoints — is the number of different ways components can be instantiated. Already we have different sources of component data:

- our own checkpoint files in safetensors
- pretrained models from huggingface
- python components configured with parameters

Now, we could simplify (1) and (2) into additional parameters, the paths to the model data. That might make it tractable.

karlhigley: Yeah, as long as the components take basic Python types as constructor parameters (like model names or string paths as you suggest), it should be easy enough to serialize the params and reconstruct the objects.

That may influence how we design components though—for example, we might have `SafeTensorsScoringModel` and `HuggingFaceScoringModel` components instead of one that does both. That doesn't necessarily mean that we need to write a component for every framework as long the ones we do provide lay down a pattern that people can follow.

src/poprox_recommender/pipeline.py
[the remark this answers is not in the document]
karlhigley: I think we can address the subtle error with better type checking of the arguments against the signature inside the pipeline code, so I'm less worried about that than you are. I'm also skeptical that people will actually read documentation even if something looks weird and/or doesn't work the way they expect. 😅 

Could we meet in the middle with automated type checking raising very verbose errors that border on documentation?

PR #57, opened 2024-07-03
Add `osx-64` as a supported dev platform

karlhigley:


mdekstrand (commented):
We can add this, but with @rburke2233 fixing his Conda install it probably isn't needed. But it also doesn't hurt anything (at this time).

PR #58, opened 2024-07-08
Add ability to update lockfile without updating poprox-concepts

mdekstrand:
This PR adds a `--freeze-concepts` option to `update-dep-lock.sh` to allow us to update the lock file (e.g. to add a new dependency) **without** also updating the version of `poprox-concepts` in use. This is useful to allow development of things that need new dependencies while someone else is working on updating the recommender to a new `poprox-concepts` version.

Adding the reverse — updating only `poprox-concepts` without other dependencies — is still blocking on conda/conda-lock#652.

### DIFF
--- a/dev/environment.yml
+++ b/dev/environment.yml
@@ -9,6 +9,7 @@ dependencies:
   - dvc >=3.51,<4
   - dvc-s3
   - conda-lock >=2.5,<3
+  - yq
   - hatch
   # dependencies for development scripts
   - docopt >=0.6
--- a/dev/update-dep-lock.sh
+++ b/dev/update-dep-lock.sh
@@ -1,12 +1,53 @@
-#!/bin/sh
+#!/bin/bash
 
 # TODO: add option to update only our dependencies once conda-lock bug is fixed
 # bug url: https://github.com/conda/conda-lock/issues/652
 
+freeze_concepts=no
+
 # make sure we're in the right directory
-if [ ! -f pyproject.toml ]; then
+if [[ ! -f pyproject.toml ]]; then
     echo "this script must be run from the project root" >&2
     exit 64
+fi
+
+while [[ -n "$1" ]]; do
+    case "$1" in
+        --freeze-concepts) freeze_concepts=yes; shift;;
+        --*)
+            echo "invalid flag $1" >&2; exit 2;;
+        *)
+            echo "unexpected argument $1" >&2; exit 2;;
+    esac
+done
+
+declare -a lock_args
+lock_args=(-f pyproject.toml -f dev/environment.yml -f dev/constraints.yml)
+
+# check if we want to freeze popropx-concepts
+if [[ $freeze_concepts = yes ]]; then
+    declare -a concept_urls
+    # extract the currently-locked POPROX URL
+    concept_urls=($(yq '.package |map(select(.name == "poprox-concepts")) | map(.url) | unique() | join("\n")' conda-lock.yml))
+    if [[ ${#concept_urls[@]} -eq 1 ]]; then
+        echo "found locked poprox-concepts at $concept_urls"
+    elif [[ ${#concept_urls[@]} -eq 0 ]]; then
+        echo "ERROR: no poprox-concepts not found in lockfile" >&2
+        exit 3
+    else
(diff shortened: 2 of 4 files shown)


dev/update-dep-lock.sh
```diff
@@ -1,19 +1,61 @@
-#!/bin/sh
+#!/bin/bash
```
mdekstrand, line 1: Updated because the new logic uses Bash arrays to deal (more) sanely with lists of arguments, etc.

dev/update-dep-lock.sh
```diff
@@ -1,19 +1,61 @@
...
 
 # TODO: add option to update only our dependencies once conda-lock bug is fixed
 # bug url: https://github.com/conda/conda-lock/issues/652
 
+freeze_concepts=no
+
 # make sure we're in the right directory
-if [ ! -f pyproject.toml ]; then
+if [[ ! -f pyproject.toml ]]; then
```
mdekstrand, line 9: Since we're `/bin/bash` now, use the more robust conditional construct.

dev/update-dep-lock.sh
```diff
@@ -1,19 +1,61 @@
...
+    elif [[ ${#concept_urls[@]} -eq 0 ]]; then
+        echo "ERROR: no poprox-concepts not found in lockfile" >&2
+        exit 3
+    else
+        echo "WARNING: lockfile has multiple poprox-concepts versions" >&2
+        for ver in "${concept_urls[@]}"; do
+            echo "- $ver" >&2
+        done
+        echo "using: ${concept_urls[0]}" >&2
+    fi
```
mdekstrand, line 43: We *should* always have exactly 1 distinct `poprox-concepts` URL (including hash) in the lockfile — I do not know of a way that would be violated in current setup. However, to make this robust in the unlikely scenario that we *do* somehow have such an invalid lock file, this is some error checking to yell about the siutation if parsing `conda-lock` does not yield exactly one source URL.

dev/update-dep-lock.sh
```diff
@@ -1,19 +1,61 @@
...
+
+    cat >dev/poprox-concepts-dynamic-constraint.yml <<EOF
+dependencies:
+- pip:
+    - ${concept_urls[0]}
+EOF
+    lock_args=("${lock_args[@]}" -f dev/poprox-concepts-dynamic-constraint.yml)
```
mdekstrand, line 50: We implement the constraint by writing an additional constraint file and including it in the `conda-lock` inputs.

dev/update-dep-lock.sh
```diff
@@ -1,19 +1,61 @@
...
 # our dependencies require flexible resolution to work on Windows, because
 # conda-forge does not ship Windows binaries for current Torch versions.
 # If users have followed the Conda recommended defaults to enable strict
 # channel priority, locking will fail, so we override that setting here.
 export CONDA_CHANNEL_PRIORITY=flexible
 
 # and run conda-lock
-exec conda-lock lock -f pyproject.toml -f dev/environment.yml -f dev/constraints.yml
+echo "+ conda-lock lock ${lock_args[*]}" >&2
+exec conda-lock lock "${lock_args[@]}"
```
mdekstrand, line 61: bit of quick debugging support. The `[*]` vs `[@]` distinction is deliberate — for the debug message, we don't want the shell to do word-splitting.

dev/environment.yml
```diff
@@ -9,6 +9,7 @@ dependencies:
   - dvc >=3.51,<4
   - dvc-s3
   - conda-lock >=2.5,<3
+  - yq
```
mdekstrand, line 12: Script updates use `yq` to parse the `conda-lock.yml` file.

mdekstrand:
There is a good argument to be made that we should be specifying versions in `pyproject.toml` instead of these shenanigans, but doing so with hashes is quite unwieldy. If we want to go that route, we should be wiring up releases and package publication into the pipeline, publishing a new numbered version (to our own repo on [AWS CodeArtifact](https://aws.amazon.com/codeartifact/)), and then depend on those artifacts instead of direct Git dependencies.

I'm happy to help build such a workflow, but doing so is outside the scope of "make it so we can add/change external deps while others are working on updates for concept changes".

PR #59, opened 2024-07-08
Update offline data generation script

sophiasun0515:


data/process_offline_test_data.py
```diff
@@ -0,0 +1,141 @@
...
+
+def load_data():
+    val_behavior_df = pd.read_table(
+        "./val_mind_large/behaviors.tsv",
+        header="infer",
+        usecols=range(5),
+        names=["impression_id", "user", "time", "clicked_news", "impressions"],
+    )
+
+    val_behavior_df.fillna("", inplace=True)
```
mdekstrand, line 26: In the interest of "minimal migration of existing logic", let's not change this yet, but going forward: I consider blanket `fillna` to be an anti-pattern — we should handle NAs in a targeted and field-appropriate manner.

data/process_offline_test_data.py
```diff
@@ -0,0 +1,141 @@
...
+    )
+
+    print(
+        f"The total unique uuid is {val_behavior_df.shape[0] + val_news_df.shape[0]}"
+    )  # The total unique uuid is 448494
+    return val_behavior_df, val_news_df
+
+
+def assign_uuid(val_behavior_df, val_news_df):
+    uuids = [uuid4() for _ in range(448494)]
```
mdekstrand, line 39: Let's not hard-code data lengths here — obtain from the appropriate place.

We can also make this deterministic, and yeet the string / uuid distinction, and also not need the length at all:

```python
ns_article = uuid5(NAMESPACE_URL, "https://data.poprox.io/mind-large/articles/")
val_news_df["uuid"] = [str(uuid5(ns_article, aid)) for aid in val_news_df["id"]]
# `str_uuid` is no longer needed
```

Behavior UUIDs can be the same — define an appropriate namespace, and use the impression ID.

mdekstrand (commented):
Couple comments. Let's at a minimum get rid of the hard-coded number, and consider doing the deterministic UUIDs (part of the idea here is that running the same script on the same input should produce the same output, unless we have a very good reason otherwise).

mdekstrand:
Let's abandon in favor of skipping the JSON file, as discussed in Slack.

PR #60, opened 2024-07-08
Add shell formatting to pre-commit

mdekstrand:
Add shfmt to pre-commit to format our shell scripts, and run it on the existing scripts.

This is an additional Go-based dependency.


mdekstrand:
Do we have the “go maybe doesn't bootstrap?” problem fixed yet?

karlhigley:
I think that turned out to be an issue with my environment, where pre-commit ran from, and which version was getting invoked

PR #61, opened 2024-07-08
Add best-effort type checking at pipeline construction and run time

karlhigley:
When components are added to the pipeline, this keeps track of the types of their inputs and outputs in order to validate that components are wired together in a way that provides the right types to the right inputs (so that e.g. flipping the order of component arguments will result in an error.) It also checks at pipeline execution time that the types being returned by components actually match their type annotations.

### DIFF
--- a/src/poprox_recommender/pipeline.py
+++ b/src/poprox_recommender/pipeline.py
@@ -76,13 +76,7 @@ class RecommendationPipeline:
 
         output = component_spec.component(*arguments)
 
-        if not isinstance(output, (ArticleSet, InterestProfile)):
-            msg = (
-                f"Pipeline components must return ArticleSet or InterestProfile, "
-                f"but received {type(output)} from {type(component_spec.component)}"
-            )
-            raise TypeError(msg)
-
+        self._validate_return_type(component_spec, output)
         state._last = component_spec.output
         state[state._last] = output
 
@@ -112,6 +106,14 @@ class RecommendationPipeline:
 
         state_type = self._state_types.get(output_name, None)
         output_type = sig.return_annotation
+
+        if output_type is not _empty and output_type not in (ArticleSet, InterestProfile):
+            msg = (
+                f"Pipeline components must return ArticleSet or InterestProfile, "
+                f"but received {type(output_type)} from {type(spec.component)}"
+            )
+            raise TypeError(msg)
+
         if state_type and output_type is not _empty and state_type != output_type:
             msg = (
                 f"Component {spec.component} returns output with type {output_type} "
@@ -121,3 +123,12 @@ class RecommendationPipeline:
             raise TypeError(msg)
 
         return output_type
+
+    def _validate_return_type(self, component_spec, output):
+        expected_type = self._state_types.get(component_spec.output, None)
+        if expected_type and not isinstance(output, expected_type):
+            msg = (
+                f"{type(component_spec.component)} is expected to return {expected_type}, "
+                f"but received {type(output)}"
+            )
+            raise TypeError(msg)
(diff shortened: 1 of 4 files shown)


karlhigley:
@mdekstrand I had to work around the way the `@torch_inference` decorator modifies the call signature types to get this to work. The workaround isn't too bad, but it seems like the kind of thing that's going to be very difficult for downstream researchers to troubleshoot on their own. Is there a slicker way to do this than what I did here?

mdekstrand:
@karlhigley I wrote that decorator in the simplest possible way, let's look for how to make decorators preserve type signatures.

karlhigley:
@mdekstrand Looks like `functools.wraps` works well 👍🏻

src/poprox_recommender/pipeline.py
```diff
@@ -71,14 +77,78 @@ def run_component(self, component_spec: ComponentSpec, state: PipelineState):
...
+def is_union(t: Any) -> bool:
+    origin = get_origin(t)
+    return origin is Union or origin is UnionType
+
+
+def included_in(object_type: type, t: Any):
+    if t == _empty or t is None:
+        return True
+
+    if object_type == t:
```
mdekstrand, line 148: Should this also have `issubclass`, possibly gated on a `covariant` boolean argument? If A subclasses B, then an output of A is valid for an input of B.

src/poprox_recommender/pipeline.py
```diff
@@ -71,14 +77,78 @@ def run_component(self, component_spec: ComponentSpec, state: PipelineState):
...
+    def _validate_output_type(self, spec: ComponentSpec):
+        # Best effort type-checking to highlight when output is
+        # mismatched with existing state values types
+        sig = signature(spec.component.__call__)
+        output_name = spec.output
+
+        state_type = self._state_types.get(output_name, None)
+        output_type = sig.return_annotation
+
+        if output_type is not _empty and output_type not in (ArticleSet, InterestProfile):
```
mdekstrand, line 111: would use `issubclass` instead of `not in`

src/poprox_recommender/pipeline.py
```diff
@@ -71,14 +77,78 @@ def run_component(self, component_spec: ComponentSpec, state: PipelineState):
...
 
-        return state
+            raise TypeError(msg)
+
+        return output_type
+
+    def _validate_return_type(self, component_spec, output):
+        expected_type = self._state_types.get(component_spec.output, None)
+        output_type = type(output)
+        if not included_in(output_type, expected_type):
```
mdekstrand, line 131: just `isinstance`? (`isinstance` can take a tuple of types, and returns `True` if it is an instance of any of the specified types.)

mdekstrand (approved):
LGTM for exact type checking. I think it would be more robust / flexible if it used `isinstance` and `issubclass` to handle input type covariance.

src/poprox_recommender/pipeline.py
karlhigley: Do we expect to subclass `ArticleSet`, `InterestProfile`, or other types we'd use this on? No objection to improving the robustness of this, just wondering how much this is a theoretical concern vs a practical problem.

src/poprox_recommender/pipeline.py
karlhigley: Same question here: do we expect to subclass those and if so why?

src/poprox_recommender/pipeline.py
mdekstrand: I think subclassing is unlikely in practice, more of a theoretical correctness concern (and a "get the type logic right now, b/c types are a place where theoretical correctness is important, and so we don't have to page this code back into working memory later if that situation changes").

PR #62, opened 2024-07-08
Update test_offline.py for current POPROX + skip MIND JSON

mdekstrand:
This PR implements Step 1 of the refactor we have discussed on Slack:

-  Move the data processing code from `test_data.ipynb` into a Python module to directly load into memory instead of saving to `mime_data.json`. I elected to put this in a separate `data/mime.py` package instead of directly in `test_offline`, and return a named tuple containing the data structures we would deserialize from disk, so we don't clutter up `test_offline` a lot. The logic is preserved as-is, except for removing the hard-coded numbering for the number of UUIDs to generate. The one change is that it loads the data directly from the MIND zip files instead of loading it from the extracted data on disk.
- Update the `test_offline` code to (mostly) work with the current recommendation pipeline. Some users fail due to mismatched data structure length; the code now logs those errors and continues.
- Add logging to better inspect what is happening in the `test_offline.py`.
- Remove the unused data files from DVC/git.

The next step is to modify so that this no longer loads everything into memory at once. I think that will be a separate PR.


src/poprox_recommender/data/mind.py
```diff
@@ -0,0 +1,189 @@
...
+class MindTables(NamedTuple):
+    """
+    News and behavior data loaded from MIND (raw data frames).
+    """
+
+    news: pd.DataFrame
+    behaviors: pd.DataFrame
+
+
+class MindData(NamedTuple):
```
karlhigley, line 33: I might use Pydantic here for consistency but no objection to moving forward with this as is

sophiasun0515 (approved):
looks good!

src/poprox_recommender/data/mind.py
mdekstrand: This is an intermediate class, it's going away in the next PR :)

PR #63, opened 2024-07-08
Incrementally process MIND data instead of pre-loading all of it

mdekstrand:
This follows up on #62 to make the offline tests process the MIND test data incrementally instead of pre-instantiating all full recommendation requests in memory.

Right now, the evaluation is unusably slow, likely due to the lack of caching of article text embeddings. The old code had a 10-article limit that resulted in unhelpful evaluations but did let the code finish.

> [!IMPORTANT]
> This also brings a correctness fix to the metrics, as the previous reciprocal rank was always 1, because it looked for *any* logged record instead of only relevant ones.

### DIFF
--- a/src/poprox_recommender/data/mind.py
+++ b/src/poprox_recommender/data/mind.py
@@ -5,133 +5,135 @@ Support for loading MIND_ data for evaluation.
 """
 
 # pyright: basic
-import json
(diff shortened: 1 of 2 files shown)


sophiasun0515 (approved):
🚢 Neat improvement!

PR #65, opened 2024-07-09
Improve documentation and type annotations for article models

mdekstrand:
This adds type annotations and improves documentation for the article embedding code pathway, and corrects seemingly-erroneous documentation for the NRMS model input.

### DIFF
--- a/src/poprox_recommender/embedders/article.py
+++ b/src/poprox_recommender/embedders/article.py
@@ -1,10 +1,26 @@
+# pyright: basic
+from typing import Protocol
+
 import torch as th
+from transformers import PreTrainedTokenizer
 
 from poprox_concepts import ArticleSet
 
 
+class ArticleEmbeddingModel(Protocol):
+    """
+    Interface exposed by article embedding models.
+    """
+
+    def get_news_vector(self, news: th.Tensor) -> th.Tensor: ...
+
+
 class ArticleEmbedder:
-    def __init__(self, model, tokenizer, device):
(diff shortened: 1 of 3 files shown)


src/poprox_recommender/embedders/article.py
```diff
@@ -1,10 +1,26 @@
...
 
 from poprox_concepts import ArticleSet
 
 
+class ArticleEmbeddingModel(Protocol):
+    """
+    Interface exposed by article embedding models.
+    """
+
+    def get_news_vector(self, news: th.Tensor) -> th.Tensor: ...
```
karlhigley, line 15: Is there anything that distinguishes this API from the model's `forward` method? We'd probably have an easier time swapping in other Torch models if we could rely on them adhering to the standard Torch model API

karlhigley: (I realize that you're codifying an existing API, but I want to avoid solidifying things that aren't where we want to end up)

mdekstrand: There is nothing at present that distinguishes it, but in the future we might not just have titles. But that isn't a reason not to name it `forward` (or to just make it a callable).

Do we want to make that change here, or keep the docs here & refactor this API as a separate future step?

karlhigley: I think this PR makes it a little harder to do the refactor, but not enough that it's worth worrying about. Let's go forward with this and leave that for later

karlhigley (approved):

PR #68, opened 2024-07-09
Add support for configuring Torch devices

mdekstrand:
This makes a couple of infrastructure tweaks to enable CUDA-based evaluation of the recomemnders:

- adds a second lockfile, `conda-lock-cuda.yml`, that locks CUDA-based Linux dependencies.
- moves the platform specification into `update-dep-lock.sh` instead of `pyproject.toml` to allow us to lock CUDA for Linux only.
- adds device configuration support through a `default_device()` function that looks for a `POPROX_REC_DEVICE` environment variable.
- runs the full evaluation and includes its results, took about 50 minutes on my 2GHz EPYC 7662 w/ an A40 (eval is currently CPU-bound).


src/poprox_recommender/pipeline.py
```diff
@@ -24,9 +26,10 @@ def __call__(self, inputs: dict[str, ArticleSet | InterestProfile]) -> ArticleSe
         # Avoid modifying the inputs
         state = deepcopy(inputs)
 
-        # Run each component in the order it was added
-        for component_spec in self.components:
-            state = self.run_component(component_spec, state)
+        with torch.inference_mode():
+            # Run each component in the order it was added
+            for component_spec in self.components:
+                state = self.run_component(component_spec, state)
```
mdekstrand, line 32: I don't know if this is the best place to put this, but we need to run the pipeline components in inference mode to avoid memory explosions.

src/poprox_recommender/model/nrms/__init__.py
```diff
@@ -22,6 +22,10 @@ def __init__(self, args):
         self.click_predictor = DotProductClickPredictor()
         self.loss_fn = nn.CrossEntropyLoss()
 
+    @property
+    def embedding_size(self) -> int:
+        return self.news_encoder.plm_hidden_size
+
```
mdekstrand, line 28: For debugging purposes, I needed a way to get the embedding size of the model.

src/poprox_recommender/embedders/article.py
```diff
@@ -34,6 +36,10 @@ def __init__(self, model: ArticleEmbeddingModel, tokenizer: PreTrainedTokenizer,
         self.embedding_cache = {}
 
     def __call__(self, article_set: ArticleSet) -> ArticleSet:
+        if not article_set.articles:
+            article_set.embeddings = th.zeros((0, self.model.embedding_size))  # type: ignore
+            return article_set
```
mdekstrand, line 41: While we don't call the profile with empty sets, this just makes this model a little more robust (although downstream components will break when this condition is triggered).

src/poprox_recommender/test_offline.py
```diff
@@ -89,25 +89,38 @@ def recsys_metric(mind_data: MindData, request: RecommendationRequest, recommend
...
+                personalized = 1
+            else:
+                recommendations = fallback(
+                    {
+                        "candidate": ArticleSet(articles=request.todays_articles),
+                        "clicked": ArticleSet(articles=request.past_articles),
+                        "profile": request.interest_profile,
+                    }
+                )
+                personalized = 0
```
mdekstrand, line 109: We should abstract this better (in-pipeline fallback support?) but for now this gets the eval working. Also record which pipeline was used for each user.

src/poprox_recommender/test_offline.py
```diff
@@ -89,25 +89,38 @@ def recsys_metric(mind_data: MindData, request: RecommendationRequest, recommend
...
         except Exception as e:
             logger.error("error recommending for user %s: %s", request.interest_profile.profile_id, e)
-            user_csv.writerow([request.interest_profile.profile_id, None, None, None])
-            nbad += 1
-            continue
+            raise e
```
mdekstrand, line 112: Decided to fix the errors instead of log them.

src/poprox_recommender/embedders/article.py
```diff
@@ -52,28 +58,33 @@ def __call__(self, article_set: ArticleSet) -> ArticleSet:
...
                 cached[uca.article_id] = a_emb
-                self.embedding_cache[uca.article_id] = a_emb
+                # use CPU for cached articles to reduce CUDA memory usage
+                self.embedding_cache[uca.article_id] = a_emb.cpu()
 
         # Step 6: stack the embeddings into a single tensor
-        embed_single_tensors = list(cached.values())
-        assert not any(e is None for e in embed_single_tensors)
+        # we do this with a list to properly deal with duplicate articles
+        embed_single_tensors = [cached[article.article_id].to(self.device) for article in article_set.articles]  # type: ignore
```
mdekstrand, line 83: Our MIND data can have duplicate articles, so this makes the embeddings match the article set's list even when there are duplicates.

src/poprox_recommender/test_offline.py
```diff
@@ -128,8 +141,8 @@ def recsys_metric(mind_data: MindData, request: RecommendationRequest, recommend
     out_fn = project_root() / "outputs" / "metrics.json"
     out_fn.parent.mkdir(exist_ok=True, parents=True)
     out_fn.write_text(json.dumps(agg_metrics) + "\n")
-    print(
-        f"Offline evaluation metrics on MIND data: NDCG@5 = {np.mean(ndcg5)}, NDCG@10 = {np.mean(ndcg10)}, MRR = {np.mean(recip_rank)}"  # noqa: E501
-    )
+    logger.info("Mean NDCG@5: %.3f", np.mean(ndcg5))
+    logger.info("Mean NDCG@10: %.3f", np.mean(ndcg10))
+    logger.info("Mean RR: %.3f", np.mean(recip_rank))
```
mdekstrand, line 147: I like loggers…

src/poprox_recommender/config.py
mdekstrand, line 1: New file! We can add additional config things here later, but for now it gives a quick way to configure compute devices with the environment.

mdekstrand (commented):
Some remarks on what this code is doing.

src/poprox_recommender/pipeline.py
karlhigley: I see the issue this solves, but I also agree that this probably isn't the best place to put this. Ideally the pipeline abstraction has no knowledge of what frameworks the components are using, maybe the way to do this is to push it down into the components that use Torch?

karlhigley (approved):

PR #69, opened 2024-07-10
Compress per-user evaluation outputs

mdekstrand:
This adds `gzip` compression to the per-user metric output CSV, bringing it from 25M to 9.4M.


karlhigley (approved):

mdekstrand:
@karlhigley I removed the subsetting so I can merge this with updated output files & metrics, and it now hits hits a problem (introduced in #74?) after the subset from initial tests:

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/mde48/poprox/poprox-recommender/src/poprox_recommender/evaluation/offline.py", line 126, in <module>
    single_ndcg5, single_ndcg10, single_rr, single_rbo5, single_rbo10 = recsys_metric(mind_data, request, outputs)
                                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mde48/poprox/poprox-recommender/src/poprox_recommender/evaluation/offline.py", line 64, in recsys_metric
    ranked = pipeline_state["ranked"]
             ~~~~~~~~~~~~~~^^^^^^^^^^
  File "/home/mde48/poprox/poprox-recommender/src/poprox_recommender/pipeline.py", line 26, in __getitem__
    return self.elements[key]
           ~~~~~~~~~~~~~^^^^^
KeyError: 'ranked'
```

karlhigley:
@mdekstrand I can reproduce the issue, so I'll dig into it. Weird that it doesn't show up with the subset applied 🤔

karlhigley:
Didn't show up with the subset applied because there weren't any users without clicks in that part of the dataset. Adjusted the output name in the fallback pipeline and made a few further tweaks to handle the case where RBO can't be computed because there aren't ranked and reranked lists to compare.

sophiasun0515 (approved):
Looks good!

PR #70, opened 2024-07-10
Apply Topic Calibration and update unit test

sophiasun0515:
Main topic calibration class cr to @e-mckinnie, I helped with the code integration and test.

### DIFF
--- a/src/poprox_recommender/diversifiers/topic_calibration.py
+++ b/src/poprox_recommender/diversifiers/topic_calibration.py
@@ -12,7 +12,10 @@ from poprox_recommender.topics import extract_general_topics, normalized_topic_c
 # topic calibration
 class TopicCalibrator:
     def __init__(self, algo_params, num_slots):
-        self.theta = float(algo_params.get("theta", 0.8))
+        # Theta term controls the score and calibration tradeoff, the higher
+        # the theta the higher the resulting recommendation will be calibrated.
+        # We choose the binary search way to start in the middle
+        self.theta = float(algo_params.get("theta", 0.5))
         self.num_slots = num_slots
 
     def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestProfile) -> ArticleSet:
--- a/src/poprox_recommender/diversifiers/topic_calibration.py
+++ b/src/poprox_recommender/diversifiers/topic_calibration.py
@@ -1,13 +1,14 @@
+from collections import defaultdict
+
 import numpy as np
 import torch as th
-from poprox_concepts import Article, ArticleSet, InterestProfile
 
-from poprox_recommender.topics import (GENERAL_TOPICS, extract_general_topics,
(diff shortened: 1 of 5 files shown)


src/poprox_recommender/diversifiers/topic_calibration.py
```diff
@@ -0,0 +1,110 @@
...
+    return normalized_topic_count(S_distr_with_candidate)
+
+
+# from https://github.com/CCRI-POPROX/poprox-recommender/blob/feature/experiment0/tests/test_calibration.ipynb
+def compute_kl_divergence(interacted_distr, reco_distr, alpha=0.01):
+    """
+    KL (p || q), the lower the better.
+
+    alpha is not really a tuning parameter, it's just there to make the
+    computation more numerically stable.
```
karlhigley, line 102: Since it isn't really a tuning parameter, we probably don't need to accept it as an argument (`topic_calibration_alpha` in the `__init__` method above

karlhigley (approved):

src/poprox_recommender/diversifiers/topic_calibration.py
e-mckinnie: I'm okay with that!

sophiasun0515: I just moved alpha to be a local var here.

sophiasun0515: I just moved alpha to be a local var here.

rburke2233:
Looks good to me.

rburke2233 (approved):

sophiasun0515:
We do have one thing uncertain about the calibration parameter. In the current code logic, theta is used to control the score and calibration tradeoff; the higher the theta, the higher the resulting recommendation will be calibrated. So if the current theta = 0.8, that means we will see 80% recommendation calibrated, with the remaining 20% accuracy (personalization).

Is that too aggressive? But in practice, we won't know the effect until we experiment with the value on ourselves and eyeball our newsletters to have an opinion.

In another example topic calibration notebook, authors set the calibration param to be 0.5 (they use lambda). So we wonder if we should also start with something more balanced or conservative (i.e. <=0.5), and iterate it for some internal experiment until we know about its power.

Welcome thoughts from the seniors! @karlhigley @rburke2233

karlhigley:
When in doubt I start in the middle and binary search from there. If 0.5 seems like it sacrifices too much relevance, then try 0.25. If it doesn't match your interests well enough, try 0.75.

rburke2233:
This is a good argument for having diversity-type metrics we can run on our off-line experiment outcomes, just so we can see in an off-line sense what the parameter tradeoffs are.

sophiasun0515:
Thank you both for the input! I changed the theta term to 0.5 so we can start testing in the middle in a binary search fashion. I think the PR is ready to be merged :)

PR #71, opened 2024-07-10
Refactor logging configuration into support function + colorize logs

mdekstrand:
This refactors the logging configuration int a `setup_logging` function (in `poprox_recommenders.logging`) to enable us to easily and consistently set up logging in any CLI tool we build in this repo, and uses this function in the offline test script.

This function also provides more robust and useful logging, including colorized output on terminals automatic log file support, along with an easy API to support `--verbose` command line options.

It also fixes logging, which was broken in #68 because Michael doesn't know how to properly use `logging.dictConfig`.

### DIFF
--- a/src/poprox_recommender/logging.py
+++ b/src/poprox_recommender/logging.py
@@ -12,22 +12,6 @@ from colorlog import ColoredFormatter
 
 logger = logging.getLogger(__name__)
 
-term_fmt = ColoredFormatter(
-    "[%(blue)s%(asctime)s%(reset)s] %(log_color)s%(levelname)8s%(reset)s %(cyan)s%(name)s %(reset)s%(message)s",  # noqa: E501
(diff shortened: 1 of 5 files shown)


src/poprox_recommender/logging.py
```diff
@@ -0,0 +1,85 @@
...
+            "WARNING": "yellow",
+            "ERROR": "red",
+            "CRITICAL": "red,bg_white",
+        },
+        secondary_log_colors={},
+        style="%",
+        # options to detect whether to colorize
+        stream=sys.stderr,
+        no_color=os.environ.get("NO_COLOR", "") != "",
+        force_color=os.environ.get("FORCE_COLOR", "") != "",
```
mdekstrand, line 68: these come from https://no-color.org/ and https://force-color.org/.

karlhigley (approved):

PR #73, opened 2024-07-15
Log the number of candidate articles and error out if it's not enough 

karlhigley:


src/poprox_recommender/handler.py
```diff
@@ -30,7 +30,13 @@ def generate_recs(event, context):
     else:
         logger.info("Using default parameters")
 
-    logger.info("Selecting articles...")
+    num_candidates = len(req.todays_articles)
```
kluver, line 33: `if num_candidates < req.num_recs` ?

karlhigley: Yeah, I guess that prevents the two article newsletter. Will update

karlhigley:
I'm inclined to merge this despite the unrelated test failure, which also needs a fix but probably shouldn't block us from adding this additional guardrail.

PR #74, opened 2024-07-15
Compute metrics like RBO over outputs from multiple pipeline stages

karlhigley:
This is a proof-of-concept for metrics that require multiple ranked sets of articles to compare. Along the way, it also organizes our evaluations and pipeline components into their own sub-packages and renames a few existing sub-packages to avoid name conflicts with public or Python built-in packages.


rburke2233:
i think I see where this is going and like the fact that it covers more types of metrics. If I understand this correctly, then I would be able to write a calibration metric because InterestProfile is now input to the recsys_metric.

karlhigley:
@rburke2233 Yeah, you could compute metrics about users from their interest profiles, and about the output of any component in the recommendation pipeline—which could include metrics computed over the raw candidate set, the article embeddings, the relevance score distribution, etc. In theory this should work for just about any metric we come up with, but in practice I'd be excited to hear about metrics that it won't cover if there some.

There's another missing concept that I've been holding off on introducing until we need it, but in addition to `ArticleSet` (candidates or recs) and `InterestProfile` we should probably also have a `RecommendationContext` (or something along those lines) to capture information like "today is Friday, July 19." That would make computing metrics w.r.t. weekly, monthly, seasonal, or yearly cycles easier, and provide a place to capture information that's important for making recommendations or computing metrics but not directly about users or items.

mdekstrand (approved):

PR #76, opened 2024-07-22
 Replace fallback pipeline with joiners like concat, fill, interleave

karlhigley:


src/poprox_recommender/default.py
```diff
@@ -83,27 +83,22 @@ def personalized_pipeline(num_slots: int, algo_params: dict[str, Any] | None = N
...
+
+    # Compute embeddings
     pipeline.add(article_embedder, inputs=["candidate"], output="candidate")
     pipeline.add(article_embedder, inputs=["clicked"], output="clicked")
     pipeline.add(user_embedder, inputs=["clicked", "profile"], output="profile")
+
+    # Score and rank articles with diversification/calibration reranking
     pipeline.add(article_scorer, inputs=["candidate", "profile"], output="candidate")
-    pipeline.add(topk_ranker, inputs=["candidate", "profile"], output="ranked")
     pipeline.add(ranker, inputs=["candidate", "profile"], output="reranked")
```
karlhigley, line 94: What's missing here is a mechanism for making the branch of the pipeline above return an empty list of articles when there are no clicks. One way to do that:
* Make the user embedding identically zero when there are no clicks
* Then the articles should all get zero scores
* Add a filter that removes articles with zero scores from the list

karlhigley: Thoughts on this @mdekstrand?

mdekstrand: At the risk of a few more `if` statements (that the typechecker can hopefully help us with), I think I would be more inclined to use `None`, and then scores be NA (Pandas handles this tolerably well, NumPy is ok as well if we just use NaN as a sentinel value, which is what Pandas does under the hood at least until it's fully Arrow-based). The advantages to this:

* fast-track null compute instead of computing a lot of zeros expensively
* no data clearly distinguished from other sources of 0
* avoid risks of "testing for exactly zero" (although filtering for `>= 1.0e-6` would fix this)

Biggest downside I see: more logic to test for sentinels.

The NaN/NA design is what LensKit uses — scorers are supposed to return NA for anything they can't score, and the fallback scorer just asks each scorer in turn to score evertything left NA by the previous ones. I would be even happier with this solution if Pandas had a proper NA like R instead of repurposing NaN, but it is the Pandas way (Pandas does have masked array types, but they aren't as widely used; raw NumPy does not have masked arrays; Arrow does).

karlhigley: Yeah, those are good points. I'll head that direction!

src/poprox_recommender/default.py
```diff
@@ -83,27 +83,22 @@ def personalized_pipeline(num_slots: int, algo_params: dict[str, Any] | None = N
...
-    sampler = UniformSampler(num_slots=num_slots)
+    # Output the plain descending-by-score ranking for comparison
+    pipeline.add(topk_ranker, inputs=["candidate", "profile"], output="ranked")
 
-    pipeline = RecommendationPipeline(name="random_topical")
+    # Fallback in case not enough articles came from the ranker
     pipeline.add(topic_filter, inputs=["candidate", "profile"], output="topical")
-    pipeline.add(sampler, inputs=["topical", "candidate"], output="ranked")
+    pipeline.add(sampler, inputs=["topical", "candidate"], output="sampled")
+    pipeline.add(fill, inputs=["reranked", "sampled"], output="recs")
```
karlhigley, line 102: Once the pipeline itself supports fallbacks, we might consider whether we want to use `Fill` or a fallback here

src/poprox_recommender/components/joiners/concat.py
```diff
@@ -0,0 +1,9 @@
+from poprox_concepts import ArticleSet
+
+
+class Concatenate:
+    def __call__(self, candidates1: ArticleSet, candidates2: ArticleSet) -> ArticleSet:
+        reverse_articles = (candidates1.articles + candidates2.articles)[::-1]
+        articles = {article.article_id: article for article in reverse_articles}
+
+        return ArticleSet(articles=list(articles.values())[::-1])
```
mdekstrand, line 9: I think I see how this code works, but I don't think it's very obvious to someone who isn't familiar with ins and outs of dictionary ordering ordering and reversing. Maybe some comments describing why we build the dictionary in reverse order, and the reverse it again, would help?

src/poprox_recommender/components/joiners/concat.py
```diff
@@ -0,0 +1,9 @@
+from poprox_concepts import ArticleSet
+
+
+class Concatenate:
+    def __call__(self, candidates1: ArticleSet, candidates2: ArticleSet) -> ArticleSet:
+        reverse_articles = (candidates1.articles + candidates2.articles)[::-1]
+        articles = {article.article_id: article for article in reverse_articles}
```
mdekstrand, line 7: Might this be more readable (less dependent on list indexing magic) if it is:

```python
articles = {article.article_id: article for article in reversed(candidates1.articles + candidates2.articles)}
```

This would also have advantage of reducing one list allocation, but that isn't a sufficient reason to do it.

src/poprox_recommender/components/joiners/concat.py
mdekstrand: for consistency with the above change, this could also be `list(reversed(articles.values()))`.

src/poprox_recommender/components/joiners/fill.py
```diff
@@ -0,0 +1,15 @@
...
+class Fill:
+    def __init__(self, num_slots):
+        self.num_slots = num_slots
+
+    def __call__(self, candidates1: ArticleSet, candidates2: ArticleSet) -> ArticleSet:
+        # Fill as many slots as you can from candidates1
+        articles = candidates1.articles[: self.num_slots]
+
+        # Fill out remaining slots from candidates2
+        articles += candidates2.articles[: self.num_slots - len(articles)]
```
mdekstrand, line 13: what happens when the index is negative here? (I do not remember that nuance of Python indexing.)

mdekstrand: For consideration — would it be easier with itertools? `list(islice(chain(candidates1.articles, candidates2.articles), self.num_slots))` — then all the edge cases around the lengths of the two article lists are handled by the library.

I don't have a strong opinion on relative readability, but making the code “obviously correct” in the face of edge cases (the key one being when `candidates1` has more than `num_slots` articles) may add complexity.

src/poprox_recommender/components/joiners/interleave.py
```diff
@@ -0,0 +1,10 @@
+from poprox_concepts import ArticleSet
+
+
+class Interleave:
+    def __call__(self, candidates1: ArticleSet, candidates2: ArticleSet) -> ArticleSet:
+        articles = [None] * (len(candidates1.articles) + len(candidates2.articles))
+        articles[::2] = candidates1.articles
+        articles[1::2] = candidates2.articles
```
mdekstrand, line 8: If you're interested in going down the itertools route, itertools doesn't have interleave but more-itertools does.

src/poprox_recommender/components/joiners/concat.py
mdekstrand: Mostly thinking that explicit use of `reversed` may be easier to understand at first look for less-experienced Python programmers — it does exactly what it says.

mdekstrand (commented):
Mostly readability + 1 edge-case correctness comment.

src/poprox_recommender/components/joiners/concat.py
karlhigley: It's interesting: I typically don't use nearly as much indexing magic on my own as the students seem to reach for

src/poprox_recommender/components/joiners/fill.py
karlhigley: I'm generally good with improving the edge case handling of this code but it's not obvious to me what the itertools snippet is actually doing without the itertools docs in front of me 😅

src/poprox_recommender/components/joiners/interleave.py
karlhigley: Of the indexing magic contained here, this is the one I think probably makes the most sense

src/poprox_recommender/components/joiners/fill.py
mdekstrand: At the minor cost of more list allocations, we could do list combination:

```python
articles = candidates1.articles + candidates2.articles
articles = articles[:self.num_slots]
```

karlhigley:
Took different approaches to improving readability in the three cases:
* For `Concatenate`, I can't think of a better way to do what I'm aiming at so I added a lengthy comment explaining what's happening
* For `Fill`, I adopted your list combination and then slicing approach
* For `Interleave`, I rewrote it with `itertools.zip_longest` which seems pretty intuitive if you know `zip`

mdekstrand (approved):

PR #78, opened 2024-07-31
Split offline test into generation and evaluation

sophiasun0515:
Convert `offline.py` into `generate.py` and `evaluate.py`; add command line support for both scripts.

Confirmed that both generation and evaluation logic is working efficiently.


src/poprox_recommender/evaluation/evaluate.py
```diff
@@ -0,0 +1,116 @@
...
+from poprox_recommender.data.mind import MindData
+from poprox_recommender.evaluation.metrics import rank_biased_overlap
+from poprox_recommender.paths import project_root
+
+logger = logging.getLogger("poprox_recommender.measurement_offline")
+
+
+def compute_rec_metric(recs_df: pd.DataFrame, request: RecommendationRequest):
+    # Make sure you run generate.py first to get the recommendations data file
+    recs = recs_df[recs_df["user"] == request.interest_profile.profile_id]
```
mdekstrand, line 21: This is the first thing to speed up. For every test user (360K+), it is doing a linear scan over the entire recs data frame, so this process is O(N^2).

If you move ``user`` to the index in the main code, before your loop:

```python
recs_df = recs_df.set_index('user', append=True)
```

Then you can do a fast lookup:

```python
recs = recs_df.loc[str(request.interest_profile.profile_id)]
```

sophiasun0515: Yeah after the change now the time is lowered to 12 hr! I'll keep exploring if there can be other optimization

mdekstrand: Yep, there definitely will be.

src/poprox_recommender/evaluation/evaluate.py
```diff
@@ -0,0 +1,116 @@
...
+    truth = mind_data.user_truth(request.interest_profile.profile_id)
+
+    personalized = 1 if request.interest_profile.click_history.article_ids else 0
+
+    single_rr = topn.recip_rank(recs[recs["stage"] == "final"], truth[truth["rating"] > 0])
+    single_ndcg5 = topn.ndcg(recs, truth, k=5)
+    single_ndcg10 = topn.ndcg(recs, truth, k=10)
```
mdekstrand, line 28: these metrics also need to filter to final.

You probabaly want to filter the recs list once (`final = recs[recs["stage"] == "final"]`).

src/poprox_recommender/evaluation/evaluate.py
```diff
@@ -0,0 +1,116 @@
...
+from poprox_recommender.evaluation.metrics import rank_biased_overlap
+from poprox_recommender.paths import project_root
+
+logger = logging.getLogger("poprox_recommender.measurement_offline")
+
+
+def compute_rec_metric(recs_df: pd.DataFrame, request: RecommendationRequest):
+    # Make sure you run generate.py first to get the recommendations data file
+    recs = recs_df[recs_df["user"] == request.interest_profile.profile_id]
+    truth = mind_data.user_truth(request.interest_profile.profile_id)
```
mdekstrand, line 22: do we remember how this function works and how slow it is?

mdekstrand:
The `line_profiler` package is also quite helpful for figuring out where the program is spending its time.

src/poprox_recommender/evaluation/evaluate.py
sophiasun0515: The mind_data.user_truth() works like this:

```
    def user_truth(self, user: UUID) -> pd.DataFrame | None:
        """
        Look up the ground-truth data for a particular user, in LensKit format,
        with item UUIDs for item IDs.
        """
        try:
            uid = self.behavior_id_for_uuid(user)
            imp_log = str(self.behavior_df.loc[uid, "impressions"])
        except KeyError:
            raise ValueError(f"unknown user {user}")

        truth = pd.DataFrame.from_records(
            ((article.split("-")[0], int(article.split("-")[1])) for article in imp_log.split()),
            columns=["mind_item_id", "rating"],
        )
        truth["item"] = [self.news_uuid_for_id(aid) for aid in truth["mind_item_id"]]
        return truth.set_index("item")
```

I remember we discussed during our split of the generation code that this piece can be saved in the evaluate; maybe we can try to run and store truth data in a separate file as well?

mdekstrand: That's probably not our bottleneck then.

mdekstrand: It's using the index so should be reasonably fast.

mdekstrand:
Got this sped up to ~30 min, and it's currently running the measurement on my server.

The key was that Pandas indexes with duplicate / hierarchical index keys are not very fast. I pre-grouped the data in to a dictionary mapping user IDs to data frames, and that lookup is virtually instantaneous.

mdekstrand:
We're still waiting on 2 things @sophiasun0515 was working on:

- [ ] make the eval script also take command-line args
- [ ] figure out why the results are wrong (new code masures NDCG = 0)

sophiasun0515:
Thanks for following up! I was working on the 0 metrics issue for the past
two days but didn’t figure out the cause yet; the eval script command-line
args is also ongoing, but I can’t guarantee to finish it by tomorrow’s
meeting (which I’ll miss due to a visa appointment, and I’m driving to
Chicago today).

I’ll update again whenever I have progress.

karlhigley:
Okay, all good! Just wanted to make sure we didn't have code ready to go that hadn't been merged (which happens more often than you'd think!)

src/poprox_recommender/evaluation/evaluate.py
```diff
@@ -0,0 +1,148 @@
...
+        single_rbo10 = rank_biased_overlap(ranked, reranked, k=10)
+    else:
+        single_rbo5 = None
+        single_rbo10 = None
+
+    return single_ndcg5, single_ndcg10, single_rr, single_rbo5, single_rbo10, personalized
+
+
+def main():
+    options = docopt(__doc__)
```
mdekstrand, line 61: This file doesn't seem to have a docstring to define the accepted options.

src/poprox_recommender/evaluation/evaluate.py
```diff
@@ -0,0 +1,148 @@
...
+    setup_logging(verbose=options["--verbose"], log_file=options["--log-file"])
+
+    global mind_data
+    mind_data = MindData()
+
+    logger.info("loading recommendations")
+    recs_fn = project_root() / "outputs" / "mind-val-recommendations.parquet"
```
mdekstrand, line 68: we should take both input and output files from the command line, so this can be reused.

mdekstrand: The command line option could just be a NAME, and if NAME is `mind-val` it expects to find `mind-val-recommendations` and produces `mind-val-user-metrics` and `mind-val-metrics`.

mdekstrand: We can leave this configurability to a next PR though.

src/poprox_recommender/evaluation/evaluate.py
```diff
@@ -0,0 +1,148 @@
...
+from tqdm import tqdm
+
+from poprox_concepts.api.recommendations import RecommendationRequest
+from poprox_concepts.domain import Article, ArticleSet
+from poprox_recommender.data.mind import MindData
+from poprox_recommender.evaluation.metrics import rank_biased_overlap
+from poprox_recommender.logging_config import setup_logging
+from poprox_recommender.paths import project_root
+
+logger = logging.getLogger("poprox_recommender.evaluate_offline")
```
mdekstrand, line 20: Let's make the logger names match the package name (`poprox_recommender.evaluation.evaluate`).

src/poprox_recommender/evaluation/generate.py
```diff
@@ -0,0 +1,139 @@
...
+from tqdm import tqdm
+
+from poprox_concepts.api.recommendations import RecommendationRequest
+from poprox_concepts.domain import ArticleSet
+from poprox_recommender.data.mind import TEST_REC_COUNT, MindData
+from poprox_recommender.default import fallback_pipeline, personalized_pipeline
+from poprox_recommender.logging_config import setup_logging
+from poprox_recommender.pipeline import PipelineState, RecommendationPipeline
+
+logger = logging.getLogger("poprox_recommender.test_offline")
```
mdekstrand, line 33: Same comment here re: logger name.

mdekstrand (approved):

src/poprox_recommender/evaluation/generate.py
```diff
@@ -0,0 +1,139 @@
...
+
+# long-term TODO:
+# - support other MIND data (test?)
+# - support our data
+
+
+def extract_recs(
```
mdekstrand, line 40: This function is no longer going to work with the new pipeline & pipeline state, so we need to update it to that.

karlhigley: @mdekstrand Since you know the new pipeline abstraction best, maybe you could add a commit here to make that adjustment?

src/poprox_recommender/evaluation/evaluate.py
sophiasun0515: Sure, let me see if I can add to this PR (sbouldn't be too much extra work)

src/poprox_recommender/evaluation/generate.py
karlhigley:
```suggestion
logger = logging.getLogger(__name__)
```

PR #81, opened 2024-08-02
Prototype using new LensKit pipeline abstraction

mdekstrand:
This is a prototype of how the new pipeline I built for LensKit (lenskit/lkpy#462) would fit into the POPROX recommenders.

### DIFF
--- a/src/poprox_recommender/default.py
+++ b/src/poprox_recommender/default.py
@@ -1,19 +1,27 @@
 # pyright: basic
 import logging
+from dataclasses import dataclass
 from typing import Any
 
-from poprox_concepts import ArticleSet, InterestProfile
+from poprox_concepts import Article, ArticleSet, InterestProfile
 from poprox_recommender.components.diversifiers import MMRDiversifier, PFARDiversifier, TopicCalibrator
 from poprox_recommender.components.embedders import ArticleEmbedder, UserEmbedder
 from poprox_recommender.components.filters import TopicFilter
 from poprox_recommender.components.rankers.topk import TopkRanker
 from poprox_recommender.components.samplers import UniformSampler
 from poprox_recommender.components.scorers import ArticleScorer
+from poprox_recommender.lkpipeline import Pipeline
 from poprox_recommender.model import get_model
-from poprox_recommender.pipeline import PipelineState, RecommendationPipeline
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
+
+
+@dataclass
+class Recommendations:
+    recs: list[Article]
+    initial: ArticleSet | None = None
+    reranked: ArticleSet | None = None
 
 
 def select_articles(
@@ -22,7 +30,7 @@ def select_articles(
     interest_profile: InterestProfile,
     num_slots: int,
     algo_params: dict[str, Any] | None = None,
-) -> PipelineState:
+) -> Recommendations:
     """
     Select articles with default recommender configuration.
     """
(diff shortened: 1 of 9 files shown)


src/poprox_recommender/default.py
```diff
@@ -1,28 +1,36 @@
...
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
 
+@dataclass
+class Recommendations:
+    recs: list[Article]
+    initial: ArticleSet | None = None
+    reranked: ArticleSet | None = None
```
karlhigley, line 24: This gives me pause because I don't think reranking is the only pipeline step that we may be interested in before and after states of. One of the advantages of `PipelineState` was that you could design your pipeline any way you wanted and still grab particular component outputs by name

src/poprox_recommender/default.py
```diff
@@ -34,16 +42,25 @@ def select_articles(
...
+
+    result = pipeline.run(*wanted, candidate=candidate_articles, clicked=clicked_articles, profile=interest_profile)
 
-    return pipeline(inputs)
+    if topk is None:
+        assert isinstance(result, ArticleSet)
+        return Recommendations(result.articles, initial=result)
+    else:
+        assert isinstance(result, tuple)
+        reranked, initial = result
```
karlhigley, line 59: Relying on tuple order here seems a potential smell—if we're going to return a data structure, it should probably be at least a `namedtuple` or `dataclass`

karlhigley: Also not wild about returning different types in different circumstances

src/poprox_recommender/default.py
```diff
@@ -82,18 +99,26 @@ def personalized_pipeline(num_slots: int, algo_params: dict[str, Any] | None = N
...
-    pipeline.add(article_embedder, inputs=["clicked"], output="clicked")
-    pipeline.add(user_embedder, inputs=["clicked", "profile"], output="profile")
-    pipeline.add(article_scorer, inputs=["candidate", "profile"], output="candidate")
-    pipeline.add(topk_ranker, inputs=["candidate", "profile"], output="ranked")
-    pipeline.add(ranker, inputs=["candidate", "profile"], output="reranked")
+    pipeline = Pipeline()
+    candidates = pipeline.create_input("candidate", ArticleSet)
+    clicked = pipeline.create_input("clicked", ArticleSet)
+    profile = pipeline.create_input("profile", InterestProfile)
+    e_cand = pipeline.add_component("embed-candidates", article_embedder, article_set=candidates)
```
karlhigley, line 106: The verb style naming here feels a little weird to me, because a component is conceptually a thing rather than an action. If you really want the verb style, I don't think `add_component` is the right method name to go with that (but I do like `add_component` because I think it aligns more naturally with the mental model that e.g. a diversifier is a component of a larger pipeline/system.)

mdekstrand: Fair! I was trying to do something consistent (and make a consistent recommendation in the docs), and wasn't 100% on this but favored “pick one, be consistent, and document the rec”.

We can use nouns / noun phrases, and it would be slightly longer in some cases but I think would read reasonably well:

```
e_cand = pipeline.add_component("candidate-embedder", ...)
```

Or if we want to name it after the thing returned:

```
e_cand = pipeline.add_component("candidate-embeddings", ...)
```

Working through that example, we have 3 possible naming conventions:

1. The name of the component itself (embedder).
2. The name of the component's return data (embeddings).
3. The action performed by the component (embed).

Based on your rationale, it sounds like you would lean towards (1).

src/poprox_recommender/default.py
mdekstrand: <sidebar> This is another place where I really wish Python's type system had named types. With named types, we could do fun things like let the input be a `TypedDict` of nodes and the output is a `TypedDict` of return data.

Right now `run` has several overloads so that its return type is statically inferred based on the node object types, when those are available (up to 5 requested node outputs).

src/poprox_recommender/default.py
mdekstrand: I've pushed an update with the component name naming.

mdekstrand: Would it make more sense to have something like `add_process` or `add_stage` and use the verb-based naming? Not sure it would, just pondering while I update the docs.

mdekstrand:
Cross-added the pipeline updates to LensKit in lenskit/lkpy#466.

mdekstrand:
Now merged and tested w/ @karlhigley's new pipeline components change.

src/poprox_recommender/default.py
karlhigley: I like the noun-style naming better because I find it easier to think about composing objects than actions, but if we went the verb-style route I was considering `add_step`

src/poprox_recommender/default.py
```diff
@@ -82,23 +85,31 @@ def personalized_pipeline(num_slots: int, algo_params: dict[str, Any] | None = N
         logger.info("Recommendations will be ranked with plain top-k.")
         ranker = topk_ranker
 
-    pipeline = RecommendationPipeline(name=diversify)
+    # TODO put pipeline name back in
```
karlhigley, line 88: Worth resolving this now or want to wait for a follow-up PR?

tests/integration/test_smoke.py
```diff
@@ -25,7 +25,7 @@ def test_direct_basic_request():
         req.num_recs,
     )
     # do we get recommendations?
-    assert len(outputs.recs) > 0
+    assert len(outputs.default.articles) > 0
```
karlhigley, line 28: Not a blocker: Wonder if this might be worth applying some syntactic sugar to avoid having to sprinkle `default` in so many places

karlhigley (approved):

src/poprox_recommender/default.py
mdekstrand: #85 adds it :)

tests/integration/test_smoke.py
mdekstrand: Quite possibly.

PR #83, opened 2024-08-07
Make all component constructor parameters readily serializable

karlhigley:
This moves model loading into the embedder components so that they can take easily serializable paths as constructor arguments instead of large binary model weights. Since we don't want to load the same model more than once in different components, it also splits the saved model into two files for the user and news encoders.

The model loading changes also make it easier to cache the pipelines than the models, which then reduces the amount of dynamic pipeline configuration we can do at request time (e.g. conditional diversifiers, various diversifier params, the number of recommendations to request.) As a result, we build a separate pipeline for each of the rankers and diversifiers we want to expose in the API.


src/poprox_recommender/components/rankers/__init__.py
```diff
@@ -1,9 +0,0 @@
-from typing import Any
-
-
-class Ranker:
-    def validate_algo_params(self, algo_params: dict[str, Any], valid_params: list[str]):
```
karlhigley, line 5: We no longer need this base class to share the validation implementation across rankers now that we've made their constructor parameters explicit (instead of a dictionary.)

src/poprox_recommender/model/nrms/__init__.py
```diff
@@ -14,11 +14,15 @@ class NRMS(torch.nn.Module):
...
         super(NRMS, self).__init__()
 
-        self.news_encoder = NewsEncoder(args)
-        self.user_encoder = UserEncoder(args)
+        self.news_encoder = NewsEncoder(
+            config.pretrained_model,
+            config.num_attention_heads,
+            config.additive_attn_hidden_dim,
+        )
+        self.user_encoder = UserEncoder(config.hidden_size, config.num_attention_heads)
```
karlhigley, line 25: We made several similar changes along these lines, where a single object bundling many values together is expanded out to list the elements/values that are actually being used. That makes it easier to see how the values flow through the code and decouples the implementation of the function/method/class from the implementation of e.g. the config.

tests/components/test_topic_calibration.py
```diff
@@ -23,13 +23,11 @@ def test_request_with_topic_calibrator():
         ArticleSet(articles=req.todays_articles),
         ArticleSet(articles=req.past_articles),
         req.interest_profile,
-        req.num_recs,
```
karlhigley, line 26: When we were building pipelines on the fly in the request/response cycle, we could configure the number of recs with a request parameter. Now that pipelines are built ahead and cached, the pipeline and component APIs as currently designed don't make that easy. We might want to make further changes to re-enable this functionality, but for now we can get away with making all recs pipelines return 10 results.

mdekstrand: IMO the easiest way to handle this will be to make the recommendation list length a runtime input instead of a configuration parameter, and wire it up as an input.

With one additional change to the new pipeline, this can even be semi-automated.

karlhigley: There are reasons why this doesn't work super-well in some cases that I encountered in the Haystack world and were really a pain. When you define a pipeline with multiple candidate sources that need to fetch more candidates than the pipeline will eventually output, making the list length dynamic causes dilemmas and headaches about how to adjust the number of retrieved candidates relative to the list length.

To make that concrete, for an eventual list length of 10 items:
* Candidate source A returns 25 items
* Candidate source B returns 25 items
Then there's some ranker that takes the combined list of candidates and picks the best ones. Now you want to evaluate a metric at different cut-offs, so the number of results returned by the last component of the pipeline needs to change.

The first hurdle you have to get over is "how do I change the number of results from all the parts of the pipeline that need to be adjusted to make that work but not the ones that don't? how do I know which ones are which?" If you change the number of eventual results returned to 5, you can do that just by truncating the list (but that not be representative of what you think you're measuring if you're still picking from 50 candidates.) If you change it to 100, then there aren't enough candidates to do that so upstream components have to get adjusted too.

Haystack ended up in a situation where you could dynamically adjust how many results came back, but it would set the number of results _from every component_ to that number, which is not at all obvious or what you'd like to have happen in this kind of scenario. That's what happens when you try to make list length dynamic without thinking carefully about it though.

karlhigley: (I'm not saying that it's an unsolvable problem but I've hesitated to make the list length dynamic on account of that unexpected complexity)

mdekstrand: That makes sense. There are some… interesting… things you can do here with the pipeline that can encode some relevant policies, I don't know if they're a *good* idea but they're fun to think about.

For example, if you want to run with "candidate lists are 5x the target rec list depth", we could do:

```python
n_recs = pipe.create_input('n_recs', int)
n_cands = pipe.add_component(multiply, a=n_recs, b=5)

cand = pipe.add_component(CandidateSelector, n=n_cands)
score = pipe.add_component(ItemScorer, items=cand)
recs = pipe.add_component(FancyRanker, items=score, n=n_recs)
```

Then `pipe.run(..., n_recs=5)` will use 25 candidates, and `pipe.run(..., n_recs=10)` will use 50.

But also hard-wiring the list length(s) makes sense; long-term it might be worth considering mechanisms to reuse unchanged components of an execution graph measuring a pipeline with different target list lengths.

src/poprox_recommender/components/embedders/article.py
```diff
@@ -24,22 +29,36 @@ class ArticleEmbeddingModel(Protocol):
...
     model: ArticleEmbeddingModel
-    tokenizer: PreTrainedTokenizer
+    tokenizer: PreTrainedTokenizer | PreTrainedTokenizerFast
     device: str | None
     embedding_cache: dict[UUID, th.Tensor]
 
-    def __init__(self, model: ArticleEmbeddingModel, tokenizer: PreTrainedTokenizer, device: str | None):
-        self.model = model
-        self.tokenizer = tokenizer
+    def __init__(self, model_path: PathLike, device: str | None):
```
mdekstrand, line 38: This will need to save the model path to a field for configurability.

src/poprox_recommender/components/embedders/user.py
```diff
@@ -1,14 +1,21 @@
...
 
 
-class UserEmbedder:
-    def __init__(self, model, device, max_clicks_per_user: int = 50):
-        self.model = model
+class NRMSUserEmbedder:
+    def __init__(self, path, device, max_clicks_per_user: int = 50):
+        config = ModelConfig()
+        self.user_encoder = UserEncoder(config.hidden_size, config.num_attention_heads)
+        checkpoint = load_file(path)
```
mdekstrand, line 14: Likewise, this will also need to save the path.

src/poprox_recommender/components/scorers/article.py
```diff
@@ -1,18 +1,17 @@
...
-        self.model = model
-
     @torch_inference
     def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestProfile) -> ArticleSet:
         candidate_embeddings = candidate_articles.embeddings
         user_embedding = interest_profile.embedding
 
         if user_embedding is not None:
-            pred = self.model.get_prediction(candidate_embeddings, user_embedding.squeeze())
+            pred = torch.matmul(candidate_embeddings, user_embedding.t()).squeeze()
```
mdekstrand, line 14: that's nice :)

src/poprox_recommender/model/nrms/news_encoder.py
```diff
@@ -9,24 +9,26 @@
 
 
 class NewsEncoder(torch.nn.Module):
-    def __init__(self, config):
+    def __init__(self, pretrained_model, num_attention_heads, additive_attn_hidden_dim):
```
mdekstrand, line 12: Need to save model path.

src/poprox_recommender/model/nrms/user_encoder.py
```diff
@@ -7,16 +7,15 @@
 
 
 class UserEncoder(torch.nn.Module):
-    def __init__(self, config):
+    def __init__(self, hidden_size, num_attention_heads):
```
mdekstrand, line 10: Need to save these parameters (or manually implement to/from config)

mdekstrand (changes requested):
Looks good! 2 things needed to make it work with the pipeline config mechanism:

- extend `AutoConfig` (included in #81) in each of the components to get the automatic `get_config` and `from_config` implementations.
- make sure all config parameters are saved as fields (I flagged all the missing ones I saw) so `get_config`'s auto-implementation can work.

Also the `build_pipeline` function will need update for merging with #81 merged.

tests/components/test_topic_calibration.py
karlhigley: Yeah, we should think carefully about this—it seems like it ought to be solvable but half-baked solutions might be worse than nothing (as we discovered evaluating Haystack pipelines.)

karlhigley:
I really don't want to extend something called `AutoConfig` on every component—not because I don't want the functionality but because _every_ component should have that functionality, which makes it an implementation detail of configurability that component authors shouldn't need to grapple with at that level of detail. Could we put an appropriately named class in between `AutoConfig` and the actual components? `Component` and `ConfigurableComponent` are both already taken or I'd reach for those as appropriate component base class names.

src/poprox_recommender/model/nrms/user_encoder.py
karlhigley: Maybe I'm missing something about how the component serialization works, but...these are Torch model classes?

karlhigley: Since the models will get loaded from disk again when the components are reconstructed from their configs, I don't get why it would matter here

mdekstrand:
@karlhigley makes sense! With the `AutoConfig` name, I was thinking in a particular Python-thinking sense of it being a mixin, but definitely see the point of making it uniform across. The design principles of the original design:

1. core component interface is specified via protocols, not inheritance
2. configurability is opt-in — no magic assumptions
3. mixins are available to aid implementation of those protocols, including automatically supporting configurability

Now, (2) might be unnecessary, either in general or in the POPROX case specifically. (1) may also be unnecessary.

As I ponder through design here, and at the risk of bikeshedding, how would you feel about `ComponentBase` being the name of the thing that components extend? That could then yield the following design that I think would be sound (pretty much what you've reviseed to here with some naming adjustments and removal of a layer):

* `ConfigurableComponent` and `TrainableComponent` remain protocols. Possibly rename them to `Configurable` and `Trainable` — there is nothing actually component-specific about them.
* Rename `AutoConfig` to `ComponentBase`, and writing a class-based component that does not extend `ComponentBase` is unusual advanced use of the pipeline.
* We don't need a new alias anymore
* The base component `Callable` type can be either `Component` or `PipelineComponent`, I don't have a strong preference there.

tests/components/test_topic_calibration.py
mdekstrand: Yeah. I'm inclined to make sure that the core abstraction doesn't paint us into too much of one corner, and we can use practical experience to refine it.

I've got another idea for making the pipelines reconfigurable too, that will be another PR:

* add a `how="reuse"` option to `clone` that reuses existing component interfaces and only clones the pipeline structure and wiring (already planning to do this)
* instantiate and cache a "default" pipeline
* pipelines with non-default list lengths can be created by calling `clone(how="reuse")` on the cached pipeline, and then using `replace_component` to replace the components that depend on list length with reconfigured ones. This depends on those components being lightweight, but I think that's the common case.

If the last step here is common enough, we could also add a `reconfigure_component` method that works like `replace_component` (types and error handling omitted for brevity):

```python
def reconfigure_component(self, node, **changes):
    comp = node.component
    cfg = comp.get_config()
    cfg.update(changes)
    c2 = comp.from_config(cfg)
    self.replace_node(node, c2, <wiring>)
```

karlhigley:
> `ConfigurableComponent` and `TrainableComponent` remain protocols. Possibly rename them to Configurable and Trainable — there is nothing actually component-specific about them.

This naming does sound better!

> Rename AutoConfig to ComponentBase, and writing a class-based component that does not extend ComponentBase is unusual advanced use of the pipeline.

Mostly fine with this rename, but I do want to emphasize that since pipelines need to be serializable that also means that components need to be serializable, which means (in the current design) that they must be configurable. I'm having a hard time figuring out when/how/why you'd want to make a component that would render your pipeline impossible to save to/load from a file. That seems like more of a bad idea than an advanced use case.

> We don't need a new alias anymore

I don't see it as just an alias—although there's only one thing mixed into it at the moment, I was viewing it as "here's the public thing you extend, which means LK can mix more things into it at will without changing the public API."

> The base component Callable type can be either Component or PipelineComponent, I don't have a strong preference there.

I don't really like either one, but I couldn't call it `Callable` and I wanted to keep a shorter name the public-facing class.

tests/components/test_topic_calibration.py
karlhigley: This has gotten complicated enough that I'm starting to be worried that either something is off about the abstractions or we're building way more functionality than we actually need 😅

mdekstrand:
Makes sense. The edge cases I see around configurable are 2:

* `Callable`s that are classes that can be instantiated with no constructor arguments. This is what the pipeline does if it encounters a component that isn't either a function or `Configurable`. This is a very unusual situation, and I don't have a concrete use case for it right now — it emerged as a side effect of making configurability a separate interface from the base component protocol + a desire to minimize constraints the pipeline places on components, so I added tests to make sure it behaved sanely, and I'm ok with it going away. I'm also ok with the possibility existing but not being documented as a viable component implementation path.
* Components whose configuration support should be implemented in some other way (e.g. passing it off to an underlying implementation from e.g. Scikit-Learn instead of duplicating fields). This can be handled fine by overriding `get_config` inherited from a base class, with no disadvantage vs. just implementing a protocol in a different way. I might even implement a `ScikitComponent` that implements the LensKit component config methods using Scikit-Learn's config tool to make it super easy to drop Scikit into pipelines, that seems useful for LensKit.

So that suggests the following refinement:

* Do the `ConfigurableComponent` -> `Configurable` (& `TrainableComponent`) rename.
* Make `Component` (or `Component`, no strong opinion here) the base class that class-based component implementations need to extend, and it has the auto-config logic that can be overridden by derived classes. The pipeline interface is documented as taking “a callable or a `Component`”, and the case where the component is neither a function nor a `Component` is either an error or has the current behavior (instantiate with default constructor args). Per your motivation or the class, future functionality can be added without breaking subclasses.
* TBD: does the logic to determine if the component can be serialized check for `Configurable` or `Component`? Checking for `Configurable` would keep current behavior.

mdekstrand:
Since this sets up the components to extend `Component`, let's merge this and #85 and then do a separate PR to clean up those interfaces (and docs / tests where needed).

tests/components/test_topic_calibration.py
mdekstrand: Yeah. Let's use the simple "bake it in" design for now and figure out how to support reconfigurability for POPROX when we actually need it.

mdekstrand (approved):

src/poprox_recommender/model/nrms/user_encoder.py
mdekstrand: You're right, false positive in my detection of configurable classes.

PR #88, opened 2024-08-13
Vectorize MMR for speedup

mdekstrand:
This is a low-hanging fruit optimization to replace nested loops in MMR with vectorized PyTorch operations. If the embeddings remain on the GPU, then this should be doing most of MMR on the GPU as well.

### DIFF
--- a/src/poprox_recommender/components/diversifiers/mmr.py
+++ b/src/poprox_recommender/components/diversifiers/mmr.py
@@ -65,4 +65,4 @@ def mmr_diversification(rewards, similarity_matrix, theta: float, topk: int):
         # step 2: find the largest value
         S[k] = torch.argmax(scores)
 
-    return S.tolist()
+    return S[S >= 0].tolist()
--- a/src/poprox_recommender/components/diversifiers/mmr.py
+++ b/src/poprox_recommender/components/diversifiers/mmr.py
@@ -2,6 +2,7 @@ import torch
 
 from poprox_concepts import ArticleSet, InterestProfile
 from poprox_recommender.lkpipeline import Component
+from poprox_recommender.pytorch.datachecks import assert_tensor_size
 from poprox_recommender.pytorch.decorators import torch_inference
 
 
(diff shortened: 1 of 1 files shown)


src/poprox_recommender/components/diversifiers/mmr.py
```diff
@@ -46,24 +42,26 @@ def mmr_diversification(rewards, similarity_matrix, theta: float, topk: int):
...
-                candidate = i
-
-        if candidate is not None:
-            S.append(candidate)
+        # find the best combo of reward and max sim to existing item
+        # first, let's pare the matrix: candidates on rows, selected items on cols
+        M = similarity_matrix[:, S]
+
+        # for each target item, we want to find the *max* simialrity to an existing.
+        # we do this by taking the min of each row.
```
karlhigley, line 50:
```suggestion
        # we do this by taking the max of each row.
```

src/poprox_recommender/components/diversifiers/mmr.py
```diff
@@ -46,24 +42,26 @@ def mmr_diversification(rewards, similarity_matrix, theta: float, topk: int):
...
+        scores = torch.max(M, axis=1)
+        assert len(scores) == len(rewards)
+
+        # now, we want to compute θ*r - (1-θ)*s. let's do that *in-place* using
+        # this scores vector. To start, multiply by θ-1 (-(1-θ)):
+        scores *= theta - 1
+
+        # with this, we can theta * rewards in-place:
+        scores.add_(rewards, theta)
```
karlhigley, line 59: Possible to name the arguments here? Not obvious how this is `theta * rewards`

mdekstrand: Not sure how much it will help readability, since their names are `other` and `alpha`, but we actually do need to name the second one — it's keyword-only.

mdekstrand: the comment should have another word, though — "we can add theta * rewards"

mdekstrand:
@karlhigley I pushed fixes to clarify the use of `add_`, and to fix several bugs (since our tests don't actually exercise MMR!).

Based on early progress timing, this fix alone looks like it will cut our eval rec generation time (for all algos!) approximately in half.

PR #90, opened 2024-08-15
Apply topic calibration even without clicks (using onboarding topics)

karlhigley:
This creates article scores from onboarding topic selections when there are no clicks for the model to use. In that case, the normal article scorer returns `None`, which is a signal to downstream components to fill in the gap if they need to.

Here the scores computation is very rough and just good enough to let topic calibration proceed as usual: the weights of onboarding topic selections that match article topics are summed and the resulting vector of article scores is divided by its max value to normalize the scores into the [0.0,1.0] range.

Adding this to topic calibration was the easiest place to start, but I think there's an argument to be made that this could/should be a fallback scorer in all the pipelines with their various rerankers/diversifiers.

### DIFF
--- a/src/poprox_recommender/components/diversifiers/topic_calibration.py
+++ b/src/poprox_recommender/components/diversifiers/topic_calibration.py
@@ -19,21 +19,16 @@ class TopicCalibrator(Component):
         self.num_slots = num_slots
 
     def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestProfile) -> ArticleSet:
-        if candidate_articles.scores is None:
-            return candidate_articles
+        normalized_topic_prefs = self.compute_topic_dist(interest_profile)
 
-        article_scores = th.sigmoid(th.tensor(candidate_articles.scores)).cpu().detach().numpy()
+        if candidate_articles.scores is not None:
+            article_scores = th.sigmoid(th.tensor(candidate_articles.scores))
+        else:
+            article_scores = self.score_articles_by_topics(candidate_articles, normalized_topic_prefs)
 
-        topic_preferences: dict[str, int] = defaultdict(int)
+        print(f"Article scores: {article_scores}")
 
-        for interest in interest_profile.onboarding_topics:
-            topic_preferences[interest.entity_name] = max(interest.preference - 1, 0)
-
-        if interest_profile.click_topic_counts:
-            for topic, click_count in interest_profile.click_topic_counts.items():
-                topic_preferences[topic] += click_count
-
-        normalized_topic_prefs = normalized_topic_count(topic_preferences)
+        article_scores = article_scores.cpu().detach().numpy()
 
         article_indices = topic_calibration(
             article_scores,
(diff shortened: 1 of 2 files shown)


sophiasun0515:
Looks good to me; do we also want to update / add news tests for this? I know there's been a bunch of refactors happened recently and realized our original `test_topic_calibration` file was gone so not too sure about this

mdekstrand (approved):

e-mckinnie:
It has a been a while since I've thought about topic calibration so I have questions which I probably knew the answer to when I wrote topic calibration but those answers no longer exist in my brain.

Isn't candidate_article.score the baseline score of the article according to the recommender? If so then it sounds like if there aren't clicks this is None? (I think I got confused because at first I thought you were trying to simulate user clicks but topic_calibration doesn't directly use click data).

I agree that making this generic for all diversifiers, or at least having it available to all diversifiers, makes the most sense, or rearranging the code in default.py.

I am very interested in what kind of output this produces. Since topic calibration is essentially measuring how similar the articles are to the users' preferences (the article score is used as well in this equation and theta controls how much of the original article score to use versus the calibration score), another option could just be changing theta to be 1 so only calibration is considered, but then all the candidate scores would be negative which might mess up how the mmr equation works.

I guess the big question is: if a user has no clicks, what kind of behavior do we want (with topic calibration)? None (currently what we're doing), roughly estimate their click history and then compare that to topic calibration (what this PR does), or just go off of their stated interests (what setting theta to 1 with maybe more tweaks would do)?

e-mckinnie:
> Looks good to me; do we also want to update / add news tests for this? I know there's been a bunch of refactors happened recently and realized our original `test_topic_calibration` file was gone so not too sure about this

It looks like it's in tests/components. But yes, can we add a test where the user profile has no clicks but an interest profile and then perform the same check (that the returned list is different than baseline)?

karlhigley:
> Looks good to me; do we also want to update / add news tests for this? I know there's been a bunch of refactors happened recently and realized our original test_topic_calibration file was gone so not too sure about this

I do think having a test for this would be good, but I wasn't entirely sure how to write one. For manual testing, I used a real request for recs sent to the recommender for a new account I created and visually checked that the recs matched the onboarding topics I supplied. That seemed to work okay, but it wasn't obvious to me how to write test assertions that mimic that process. I suppose we could check that the item with the most matching tags has the highest score?

karlhigley:
> I am very interested in what kind of output this produces. Since topic calibration is essentially measuring how similar the articles are to the users' preferences (the article score is used as well in this equation and theta controls how much of the original article score to use versus the calibration score), another option could just be changing theta to be 1 so only calibration is considered, but then all the candidate scores would be negative which might mess up how the mmr equation works.

This was the first thing I tried but in order for it to work the scores have be floats (and not `None`.) My next thought was to make all the scores `0.0` so that only calibration counts, but that creates an issue with picking the first article because when the scores are all identical `argmax()` always returns `0` so it always picks the first article in the candidate list which might not be relevant at all.

> I guess the big question is: if a user has no clicks, what kind of behavior do we want (with topic calibration)? None (currently what we're doing), roughly estimate their click history and then compare that to topic calibration (what this PR does), or just go off of their stated interests (what setting theta to 1 with maybe more tweaks would do)?

A different way to approach this would be to adjust the way topic calibration is implemented so that it doesn't pick the first article based on the relevance score alone but instead runs through the calibration process for all positions in the list (including the first.) That's probably a simpler solution so maybe that's the way to go.

karlhigley:
I took the `zeros` approach which is largely covered by the existing tests. I'm gonna go ahead and merge in order to get this deployed before we start seeing sign-ups from Friends of Poprox (e.g. the advisory board.)

PR #91, opened 2024-08-21
Rename article `title` and `content` to `headline` and `subhead`

karlhigley:
These names probably originated with the early articles we scraped directly from AP's website, which were stored with different field names than are commonly used in journalism. Journalists sometimes call these the `hed` and `subhed`/`dek`, so I've adapted those names a bit to be easier for us to use while staying as true to the source domain as I can.

This differentiates the `subhead` (formerly `content`) from the actual text of the article, which we may soon start fetching from AP with additional API calls.

Connected PRs:
* https://github.com/CCRI-POPROX/poprox-concepts/pull/22
* https://github.com/CCRI-POPROX/poprox-storage/pull/40
* https://github.com/CCRI-POPROX/poprox-platform/pull/142
* https://github.com/CCRI-POPROX/poprox-recommender/pull/91

### DIFF
--- a/src/poprox_recommender/data/mind.py
+++ b/src/poprox_recommender/data/mind.py
@@ -123,9 +123,8 @@ class MindData:
         article = Article(
             article_id=uuid,
             url=f"urn:uuid:{uuid}",
-            title=str(self.news_df.loc[id, "title"]),
+            headline=str(self.news_df.loc[id, "title"]),
         )
-        article.content = article.title
         return article
 
 
--- a/conda-lock-cuda.yml
+++ b/conda-lock-cuda.yml
@@ -15,7 +15,7 @@
 version: 1
 metadata:
   content_hash:
-    linux-64: 07ff711fff4b49c0ca6c6c1cfc287b715ca593e4dc2fa315dd2d3e82d6687d9b
+    linux-64: 6c8a70cb28b4af58259102245eb79910bef3daae1a09066a9ac84adadc1351c1
   channels:
   - url: pytorch
     used_env_vars: []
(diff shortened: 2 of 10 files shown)


src/poprox_recommender/data/mind.py
```diff
@@ -123,9 +123,8 @@ def lookup_article(self, *, id: str | None = None, uuid: UUID | None = None):
         article = Article(
             article_id=uuid,
             url=f"urn:uuid:{uuid}",
-            title=str(self.news_df.loc[id, "title"]),
+            headline=str(self.news_df.loc[id, "title"]),
         )
-        article.content = article.title
```
karlhigley, line 128: If I updated this line it would read:
```python
article.subhead = article.headline
```
which doesn't seem right to me. If there isn't a subhead on these articles I don't think that's a problem for either the recommender or the evaluation. (Maybe I'm wrong though?)

karlhigley:
(Tests won't pass until the `concepts` PR is merged and the lock file is updated)

src/poprox_recommender/data/mind.py
kluver: According to: https://github.com/msnews/msnews.github.io/blob/master/assets/doc/introduction.md#newstsv
There should be a field `Abstract`
But I don't know much about MIND or have a good way to check this.

@sophiasun0515?

kluver (approved):

sophiasun0515:
> According to: https://github.com/msnews/msnews.github.io/blob/master/assets/doc/introduction.md#newstsv
> There should be a field Abstract
> But I don't know much about MIND or have a good way to check this.

Yep the `Abstract` field looks good, and that's also what we have from the mind data in offline evaluation script. We've wrapped it into the `[news_df](https://github.com/CCRI-POPROX/poprox-recommender/blob/main/src/poprox_recommender/data/mind.py#L30)` field; that we can access.

karlhigley:
I don't see anything in [load_mind_frames](https://github.com/CCRI-POPROX/poprox-recommender/blob/main/src/poprox_recommender/data/mind.py#L132) that indicates that we're loading the abstract field.

PR #93, opened 2024-08-28
Replace fallbacks with lazy nodes and components

mdekstrand:
This addresses a question early in discussion on fallbacks where @karlhigley suggested that a fallback should be a component. The way I see to do that without excessive recomputation is through lazy node inputs.

This introduces the following changes:

1.  Add a `Lazy` type so that a component can take a `Lazy[T]` instead of `T`, and the pipeline will key off that type declaration to defer computation of that input until it is requested.
2. Reimplement the fallback tests to use lazy nodes.
3. Update the runner to defer computation of lazy inputs.
4. Update the `Fill` POPROX component to take its second input lazily, and only request it if the first set of recommendations is too short.
5. Update the combiner tests to use the new pipeline.
6. Yeet the `use_first_of` method, `FallbackNode`, and related obsolete logic.

The result is a reduction in the number of concepts in the pipeline — rather than components, and edges, and other weird things like fallbacks, we just have components and edges (and literals).

It builds on #92, and will be easier to review once that is merged.

### DIFF
--- a/src/poprox_recommender/components/joiners/fill.py
+++ b/src/poprox_recommender/components/joiners/fill.py
@@ -1,11 +1,15 @@
 from poprox_concepts import ArticleSet
 from poprox_recommender.lkpipeline import Component
+from poprox_recommender.lkpipeline.types import Lazy
 
 
 class Fill(Component):
     def __init__(self, num_slots):
         self.num_slots = num_slots
 
-    def __call__(self, candidates1: ArticleSet, candidates2: ArticleSet) -> ArticleSet:
-        articles = candidates1.articles + candidates2.articles
+    def __call__(self, candidates1: ArticleSet, candidates2: Lazy[ArticleSet]) -> ArticleSet:
+        articles = candidates1.articles
+        if len(articles) < self.num_slots:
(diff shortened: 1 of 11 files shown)


src/poprox_recommender/components/joiners/fill.py
```diff
@@ -1,11 +1,15 @@
...
 class Fill(Component):
     def __init__(self, num_slots):
         self.num_slots = num_slots
 
-    def __call__(self, candidates1: ArticleSet, candidates2: ArticleSet) -> ArticleSet:
-        articles = candidates1.articles + candidates2.articles
+    def __call__(self, candidates1: ArticleSet, candidates2: Lazy[ArticleSet]) -> ArticleSet:
+        articles = candidates1.articles
+        if len(articles) < self.num_slots:
+            articles = articles + candidates2.get().articles
```
mdekstrand, line 13: This is the key bit that this change enables for current POPROX components — `Fill`'s second-set upstream is only executed if the first branch fails to return enough nodes.

karlhigley (approved):

PR #94, opened 2024-09-06
Add a pipeline that reranks with a softmax sampler over the article scores

karlhigley:
This add a new `SoftmaxSampler` component and creates a recommendation pipeline that applies it as the reranker. Since we have a framework for gluing component together and our recommenders have a lot of overlap, swapping it in as the reranker only takes a few lines of code beyond what's required to implement the component itself.


sophiasun0515 (approved):
lgtm!

PR #97, opened 2024-09-19
Update batch-eval logic and timings

mdekstrand:
This updates the timing information for running the batch evaluation, along with the actual evaluation results, with a few eval infra changes:

1. the generated recommendations now include a `rank` column.
2. `measure` switched from `concurrent.futures.ProcessPoolExecutor` to `ipyparallel` to work around a deadlock problem multiple of us were seeing on Macs.


mdekstrand:
Evaluations with these changes are currently running to commit the updated `dvc.lock` files before merging.

sophiasun0515 (approved):
Looks good!

karlhigley (approved):
Looks good! Since this takes a while, we might want to consider adding an option for a smaller dataset.

mdekstrand:
@karlhigley Yes, I think adding an option to subset the eval set would be good; we'll need to think how we want that to interact with the full reproduction pipeline, but the ability to get initial eval in less time is a good idea.

outputs/mind-val-metrics.csv
```diff
@@ -1,6 +1,6 @@
...
-mmr,0.3600005430536384,0.5048719979180333,0.3127684301069502,0.2272204852990004,0.32333596310501855
-nrms,0.4519876731521633,0.5470178037046607,0.3715062139527769,0.39126863333961964,0.6078344546963851
-pfar,0.4532153230343001,0.5490853194907499,0.3715732620565264,0.39126769996626554,0.6078332309671928
-softmax,0.4185689050229185,0.5174518536614011,0.3331735706112484,0.20993876957587698,0.37264024427514536
-topic-cali,0.4531301007763522,0.5492150671208499,0.3715732620565264,0.39126769996626554,0.6078332309671928
+mmr,0.35999685966956646,0.5048645294952324,0.3127684301069502,0.2272204852990004,0.32333596310501855
+nrms,0.45195517627733767,0.5471667780067349,0.3715114031212084,0.39126863333961964,0.6078344546963851
+pfar,0.45328565914654956,0.5491957605190021,0.3715732620565264,0.39126769996626554,0.6078332309671928
+softmax,0.41823221301743035,0.5172591078148666,0.33334367372380924,0.21036475800792087,0.3730802492433327
+topic-cali,0.4532503511257382,0.5491225605864557,0.3715732620565264,0.39126769996626554,0.6078332309671928
```
karlhigley, line 6: These are interesting results, given what we know about the subjective experience of receiving these recs. Makes me wonder if the topic calibration is actually doing much on the MIND data—I suspect it might not be since the topics are probably different in that data.

PR #103, opened 2024-09-23
News Locality Calibration

sophiasun0515:
Apply some refactoring by creating a `Calibrator` object that both `TopicCalibration` and the new `LocalityCalibration` inherit from. Updated current `test_calibration` test suites.

Some TODOs we need to figure out for next step:

- [x] Add test for current logic
- [ ] Tune the parameter of calibration (The current params are borrowed from topic calibration)
- [ ] Run the code on an S3 instance and call necessary endpoints from POPROX
- [ ] Figure out how to connect the logic to participant selection (who will be included in this experiment?) 
- [ ] Integrate the LLM context generation into calibrated articles


src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -0,0 +1,76 @@
...
+        article_indices = self.calibration(
+            article_scores,
+            candidate_articles.articles,
+            normalized_locality_prefs,
+            self.theta,
+            topk=self.num_slots,
+        )
+        return ArticleSet(articles=[candidate_articles.articles[int(idx)] for idx in article_indices])
+
+    def calibration(self, relevance_scores, articles, preferences, theta, topk) -> list[Article]:
```
karlhigley, line 37: This method looks pretty much the same as the corresponding method on the `TopicCalibrator`. Wonder if it could move into the base class?

src/poprox_recommender/recommenders.py
```diff
@@ -32,7 +36,7 @@ def select_articles(
     pipeline state whose ``default`` is the final list of recommendations.
     """
     available_pipelines = recommendation_pipelines(device=default_device())
-    pipeline = available_pipelines["nrms"]
+    pipeline = available_pipelines["cali"]
```
karlhigley, line 39: This will change the default when no recommender is specified in the request from plain NRMS to NRMS+Locality calibration. That's probably okay if you plan to run this experiment from a separate endpoint, but maybe not something we want to do in the main/default recommender endpoint

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -0,0 +1,76 @@
...
+                    best_candidate_score = adjusted_candidate_score
+                    candidate = article_idx
+
+            if candidate is not None:
+                recommendations.append(candidate)
+                self.add_article_to_categories(rec_categories, articles[candidate])
+
+        return recommendations
+
+    def normalized_categories_with_candidate(self, rec_categories, article):
```
karlhigley, line 68: This looks very similar to the method on the other class too and might also be able to move to the base class

karlhigley (commented):
Looks good!

src/poprox_recommender/topics.py
```diff
@@ -43,6 +76,19 @@ def user_topic_preference(past_articles: list[Article], click_history: list[Clic
     return topic_count_dict
 
 
+def user_locality_preference(past_articles: list[Article], click_history: list[Click]) -> dict[str, int]:
```
zentavious, line 79: Do we want to calibrate on the user's locality preference? I thought we might want to calibrate on the distribution of today's articles or a fixed distribution from our prior investigation.

Notably, I see an issue at the cold start of our experiment where users might have a very focused click history due to the focus of the baseline POPROX algorithm and their locality preference will be very biases.

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -0,0 +1,76 @@
...
+                    rec_categories, articles[article_idx]
+                )
+                calibration = compute_kl_divergence(preferences, normalized_candidate_topics)
+
+                adjusted_candidate_score = (1 - theta) * article_score - (theta * calibration)
+                if adjusted_candidate_score > best_candidate_score:
+                    best_candidate_score = adjusted_candidate_score
+                    candidate = article_idx
+
+            if candidate is not None:
```
zentavious, line 62: Is the only case when candidate would be None if we've already appended the entire list of article_idx to recommendations?

zentavious: Asking because if a downstream class expects there to always be 10 recommendations this could be an ideal spot to throw an error or at least a warning (not sure how much logging POPROX has or wants to have)

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -0,0 +1,76 @@
...
+            topk=self.num_slots,
+        )
+        return ArticleSet(articles=[candidate_articles.articles[int(idx)] for idx in article_indices])
+
+    def calibration(self, relevance_scores, articles, preferences, theta, topk) -> list[Article]:
+        # MR_i = \theta * reward_i - (1 - \theta)*C(S + i) # C is calibration
+        # R is all candidates (not selected yet)
+
+        recommendations = []  # final recommendation (topk index)
+        rec_categories = defaultdict(int)  # frequency distribution of catregories of S
```
zentavious, line 42: Not sure if I understand this comment. Maybe its just copied from topic_calibration?

src/poprox_recommender/recommenders.py
zentavious: This raises a good point. Are we expected to fork the repo and run the fork? 

Is POPROX planning to have a release cycle so we can uptake changes to the main POPROX?

src/poprox_recommender/components/diversifiers/locality_calibration.py
zentavious, line 37: Request: This function could also return a list of flags for articles that have been swapped from the original relevance_scores list for easy manipulation of those articles' slugs in a later stage.

zentavious (changes requested):
Only major point we should talk about is calibrating on user's locality preferences vs. a fixed locality distribution.

Otherwise everything else is primarily clarifying questions.

src/poprox_recommender/components/diversifiers/locality_calibration.py
karlhigley: Our intent is to make the downstream code robust in the sense that the expected number of recommendations will always be generated if you construct your recommendation pipeline to do so. That doesn't imply that this code needs to always produce 10 recommendations though, because you can (for example) use the `Fill` component downstream to fill in however many slots are left with a second source of recommendations (like the `UniformSampler`, which chooses at random.)

src/poprox_recommender/recommenders.py
karlhigley: Current plan is to have people fork the repo and deploy from the fork. No current plans for a defined release cycle since it hasn't become an issue yet.

src/poprox_recommender/components/diversifiers/locality_calibration.py
karlhigley: I think the `S` vs `R` nomenclature comes from the MMR paper, which then got carried into our calibrator implementation. `R` is the set of available candidates Remaining; `S` is the set of items Selected so far.

src/poprox_recommender/components/diversifiers/locality_calibration.py
karlhigley: Relevance scores is sorted in a sense, but not in order descending by score. The order matches the list of candidate articles, so that the same position in the list of articles and in the list of scores correspond to each other.

In theory, this calibration component could compute the top-k list and do some kind of comparison to generate the kind of flags you're talking about, but it currently generates its own list of recs directly from the scores rather than operating on the top-k list and swapping things around.

If you want to identify which positions of the list are different from the plain top-k list, I'd suggest using this component, [the TopkRanker component](https://github.com/CCRI-POPROX/poprox-recommender/blob/main/src/poprox_recommender/components/rankers/topk.py), and comparing the output of the two in a new `GenerateExplanations` component.

sophiasun0515: Yeah that's certainly possible, I was bit hesitant before since the `calibration` function will call overridden `add_article_to_categories()` function, but maybe that's OK as well. I'll try factor `calibration()` and `normalized_categories_with_candidate()` into the parent class and see how it goes!

src/poprox_recommender/recommenders.py
sophiasun0515: Certainly! So based on the conversation looks like it's OK to keep the pipeline as `cali` instead of `nrms` if we will run only run this with allocated participants?

I'm actually also thinking about the possibility of running some experiment flag to determine which pipeline do we want to choose here:

```
if interest_profile.experiment_id == LOCALITY_EXPT_ID:
    pipeline = available_pipelines["cali"]
else:
    pipeline = available_pipelines["nrms"]
```

But I realized that might not be the way we want to allocate users based on experiment manifest?

src/poprox_recommender/topics.py
sophiasun0515: Could you elaborate on what we might want to calibrate on in this function? I was thinking it as just a function to compute the "groud_truth" of user click history with the system, and the actual calibration step happened after we get their preference (in the new `LocalityCalibration` class). 

The cold-start issue might exist, but if we have an one-month window to observe users and collect their preferences, can that be a feasible solution?

src/poprox_recommender/components/diversifiers/locality_calibration.py
sophiasun0515: Thanks for pointing us to the `TopRanker` component; Indeed there should be a new `GenerateExplanations` component where we'll apply the LLM logic. We'll discuss that with the team.

src/poprox_recommender/recommenders.py
karlhigley: By the time the endpoint defined in this repo receives a request, it will already be for the appropriate recommender for each user according to the group assignments. The section that defines the experiment recommenders in the manifest will look something like this:
```toml
[recommenders.calibration]
endpoint = "https://n07svs3qeh.execute-api.us-east-1.amazonaws.com/?pipeline=calibration"
```
The platform side sends requests for users assigned to groups that use that recommender to that URL, and the query parameter at the end of the URL gets used by [this code](https://github.com/CCRI-POPROX/poprox-recommender/blob/main/src/poprox_recommender/recommenders.py#L34-L39) to select the appropriate pipeline.

(It's set up that way so that the recommender code doesn't need to know anything about experiments, groups, phases, etc in order to keep it simple for experimenters.)

karlhigley (approved):

PR #107, opened 2024-10-01
Fix test xfail logic

mdekstrand:
The `mark.xfail` marker did not do what I thought it would do — if `condition` was true, it didn't even try to run the test, when I wanted it to only allow the failure to be expected if the condition was true.

This PR replaces `mark.xfail` with exception-catching and imperative `xfail`, so that it does try to run the tests in all cases.


sophiasun0515 (approved):
Looks good!

PR #111, opened 2024-10-14
Plots for metrics in `mind-val-metrics.csv`

sushobhan2024:
This PR adds a folder `notebooks` and a first notebook visualizing metrics in `mind-val-metrics.csv`.


notebooks/MINDEvalResults.md
```diff
@@ -0,0 +1,140 @@
...
+from scipy.stats import sem, t
+from IPython.display import display, HTML
+
+# Loading Data
+mind_val_user_metrics = pd.read_csv('../outputs/mind-val-user-metrics.csv.gz')
+mind_val_user_metrics.head()
+```
+
+## 2. Visualizations
+To evaluate the plot, look for higher values in the bars. Higher values indicate better performance in terms of the respective metric. Error bars show the variability of the results. The exact values of the error bars are provided in the summary table below.
```
mdekstrand, line 39: This is not always true — RBO isn't “better”, it's just documenting a change. For our effectiveness metrics, higher is better.

Let's put the interpretation notes with each metric family.

notebooks/MINDEvalResults.md
```diff
@@ -0,0 +1,140 @@
...
+import pandas as pd
+import matplotlib.pyplot as plt
+import seaborn as sns
+from scipy.stats import sem, t
+from IPython.display import display, HTML
+
+# Loading Data
+mind_val_user_metrics = pd.read_csv('../outputs/mind-val-user-metrics.csv.gz')
+mind_val_user_metrics.head()
+```
```
mdekstrand, line 36: Style issue, but I would separate out imports from data load, and make the separate sub-sections with any descriptions in the notebook text instead of in code comments.

notebooks/MINDEvalResults.md
```diff
@@ -0,0 +1,140 @@
...
+        ),
+        axis=1
+    )
+
+    summary_table = summary_table[['recommender', 'Mean (95% CI)']]
+
+    html_tables+= f"<div style='display: inline-block; margin-right: 50px;'><h3>Summary Table for {metric}</h3>{summary_table.to_html(index=False, escape=False)}</div>"
+
+display(HTML(html_tables))
+```
```
mdekstrand, line 140: I like where this is going to summarize the results.

Style-wise, it would be nice to factor these kinds of things out into library code in the `poprox_recommender` package, so we don't have to repeat it throughout different notebooks.

We might want to use my `eval_tables` code, at least as a starting point — I send you an invite to the repository where that is currently living, and we can vendor it into `poprox_recommender ` for now if we want to.

mdekstrand (commented):
I've made a few comments, some on style but one on accuracy of interpretation and where interpretation guidance should go.

notebooks/notebooks.yaml
```diff
@@ -0,0 +1,14 @@
+stages:
+  results-mind-val:
+    cmd: jupytext --to notebook --execute mind-val.md
+    deps:
+      - notebooks
```
mdekstrand, line 5: The dependencies need to be:

- the markdown file
- the input metrics file

All paths need to be relative to this directory, not project root.

notebooks/notebooks.yaml
```diff
@@ -0,0 +1,14 @@
...
+    outs:
+      - notebooks/mind-val.ipynb
```
mdekstrand, line 7: Needs to be relative to this directory, not project root.

Also, let's make this output _uncached_ so we can browse the notebooks on GitHub:

```
- mind-val.ipynb:
    cache: false
```

notebooks/notebooks.yaml
```diff
@@ -0,0 +1,14 @@
+stages:
```
mdekstrand, line 1: as we discussed offline, this needs to be `dvc.yaml`.

mdekstrand (commented):
a few quick comments to fix the dvc stages

mdekstrand (approved):
would be good to commit the `dvc.lock` that results from reproducing the notebooks, but we're good to merge.

PR #116, opened 2024-10-23
Starting point for LLM article summaries

2berry2:
Uses FalconsAI model to generate summary of article and plug it as subheading


karlhigley:
Superseded by #121

PR #117, opened 2024-10-24
Update pre-commit and copier template with secret scanning

mdekstrand:
This adds [gitleaks][] to the Copier template and imports those changes here, updates our other pre-commit hook versions, and replaces our `local` hooks with externally-maintained hooks that have emerged since last setup.

This will reduce the risk of accidentally committing secrets to the repository.

[gitleaks]: https://github.com/gitleaks/gitleaks

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,14 +39,7 @@ repos:
       # Run the formatter.
       - id: ruff-format
         name: format Python source
-  - repo: local
+  - repo: https://github.com/scop/pre-commit-shfmt
+    rev: v3.10.0-1
     hooks:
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -14,15 +14,19 @@ repos:
       - id: end-of-file-fixer
       - id: mixed-line-ending
       - id: trailing-whitespace
+  - repo: https://github.com/gitleaks/gitleaks
+    rev: v8.21.1
+    hooks:
+      - id: gitleaks
   - repo: https://github.com/google/yamlfmt
```
mdekstrand, line 21: This is the key bit that gives us secret scanning.

karlhigley (approved):

PR #118, opened 2024-10-28
Add category and subcategory mentions to articles from MIND data

karlhigley:


karlhigley:
@mdekstrand I think this will do what we need, but I'm not sure how to test this

mdekstrand:
If the MIND data only has one category or subcategory per item, I think this will work.

Running `dvc repro -f measure-mind-small` will ensure that it didn't break anything currently working, but that won't directly test output correctness, and will take a little while.

mdekstrand:
Faster test:

```
python -m poprox_recommender.evaluation.generate -M MIND_small_dev -o ouptuts/mind-small-subset-recommendations.parquet --subset=100
```

karlhigley:
> If the MIND data only has one category or subcategory per item, I think this will work.

Indeed, they are represented as single value columns in the original TSV data files

karlhigley:
The fast test that it doesn't break anything runs okay, so at the very least this doesn't seem to hurt anything

PR #119, opened 2024-10-28
Create an `EmbeddingCopier` component

karlhigley:
This copies embeddings from a candidate set to an selected set of articles, so that embeddings can be used in downstream analysis for metrics like ILD.


karlhigley:
"Do we?" is Ekstrand for "We do," so I'll go fix that 😃

karlhigley:
@mdekstrand Based on our conversation yesterday, I think the conclusion was that this should work okay as long as the candidate set input has embeddings?

mdekstrand (approved):
Would like to see some docstrings that emphasize what's needed for this, but otherwise looks good to go.

karlhigley:
I'll write some docstrings and error handling before merging this

PR #120, opened 2024-10-28
add locality calibration and text generation

imrecommender:
**Description:**

This PR introduces the basic framework for integrating locality calibration and text generation into the recommender system for the locality experiment. The main components are as follows:

Locality Calibration: Calibrates recommendations based on today’s local news distribution to align with relevant geographical content.
Text Generation: Generates personalized narratives for recommended articles by referencing similar articles previously read by the user.

**To-Do:**
1. Determine the optimal number of clicked articles to generate each narrative.
2. Decide whether to implement a cold start handling mechanism for narrative generation.
3. Confirm if clicked articles should be sorted chronologically forward or backward for narrative accuracy.

### DIFF
--- a/src/poprox_recommender/components/diversifiers/locality_calibration.py
+++ b/src/poprox_recommender/components/diversifiers/locality_calibration.py
@@ -1,19 +1,41 @@
+from collections import defaultdict
+
+import numpy as np
 import torch as th
+from openai import OpenAI
+from sentence_transformers import SentenceTransformer
+from sklearn.metrics.pairwise import cosine_similarity
 
-from poprox_concepts import ArticleSet, InterestProfile
-from poprox_recommender.components.diversifiers.calibration import Calibrator
-from poprox_recommender.topics import extract_locality, normalized_category_count
+from poprox_concepts import Article, ArticleSet, InterestProfile
+from poprox_recommender.lkpipeline import Component
+from poprox_recommender.topics import extract_general_topics, extract_locality, normalized_category_count
 
 
-# Locality Calibration uses MMR
-# to rerank recommendations according to
-# locality calibration
-class LocalityCalibrator(Calibrator):
-    def __init__(self, theta: float = 0.1, num_slots=10):
-        super().__init__(theta, num_slots)
+def compute_kl_divergence(interacted_distr, reco_distr, kl_div=0.0, alpha=0.01):
+    """
+    KL (p || q), the lower the better.
+
+    alpha is not really a tuning parameter, it's just there to make the
+    computation more numerically stable.
+    """
+    for category, score in interacted_distr.items():
+        reco_score = reco_distr.get(category, 0.0)
+        reco_score = (1 - alpha) * reco_score + alpha * score
+        if reco_score != 0.0:
+            kl_div += score * np.log2(score / reco_score)
+
+    return kl_div
+
+
+class LocalityCalibrator(Component):
+    def __init__(self, theta_local: float = 0.1, theta_topic: float = 0.1, num_slots=10):
+        self.theta_local = theta_local
+        self.theta_topic = theta_topic
+        self.num_slots = num_slots
 
     def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestProfile) -> ArticleSet:
-        normalized_locality_prefs = normalized_category_count(interest_profile.click_locality_counts)
+        normalized_topic_prefs = self.compute_topic_dist(interest_profile)
+        normalized_locality_prefs = self.compute_local_dist(candidate_articles)
 
         if candidate_articles.scores is not None:
             article_scores = th.sigmoid(th.tensor(candidate_articles.scores))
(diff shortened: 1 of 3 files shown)


src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -25,13 +47,234 @@ def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestPro
...
+        rec_categories_with_candidate = rec_categories.copy()
+        self.add_article_to_categories(rec_categories_with_candidate, article)
+        return normalized_category_count(rec_categories_with_candidate)
+
+    def normalized_localities_with_candidate(self, rec_localities, article):
+        rec_localities_with_candidate = rec_localities.copy()
+        self.add_article_to_localities(rec_localities_with_candidate, article)
+        return normalized_category_count(rec_localities_with_candidate)
+
+    def calibration(
```
karlhigley, line 80: Overriding this method to calibrate across the two specific dimensions you care about for this study is an expedient way to get where you want to go and probably makes sense here. Outside the scope of this PR and not a requested change to this code, but we should also think about how to make the default implementation in the base class flexible enough to handle an arbitrary number of dimensions.

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -25,13 +47,234 @@ def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestPro
...
-        locality_list = extract_locality(article)
-        for locality in locality_list:
-            rec_categories[locality] = rec_categories.get(locality, 0) + 1
+        recommended_list = text_generation(candidate_articles.articles, article_indices, interest_profile)
+        return ArticleSet(articles=recommended_list)
+
+    def add_article_to_categories(self, rec_topics, article):
+        topics = extract_general_topics(article)
+        for topic in topics:
+            rec_topics[topic] = rec_topics.get(topic, 0) + 1
```
karlhigley, line 63: Since the topics are `defaultdict`s, `rec_topics[topic]` will return 0 even `topic` isn't present in the dictionary yet, so here you can actually just do:
```suggestion
            rec_topics[topic] += 1
```
The same thing is true for localities in the method below

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -25,13 +47,234 @@ def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestPro
...
+
+    def add_article_to_localities(self, rec_localities, article):
+        localities = extract_locality(article)
+        for local in localities:
+            rec_localities[local] = rec_localities.get(local, 0) + 1
```
karlhigley, line 68: These methods that modify their parameters should work fine but are a little awkward compared to the methods below that return a value. Maybe the two methods above could also copy the parameter, make modifications, and return the modified copy?

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -25,13 +47,234 @@ def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestPro
...
+                    candidate = article_idx
+
+            if candidate is not None:
+                recommendations.append(candidate)
+                self.add_article_to_categories(topic_categories, articles[candidate])
+                self.add_article_to_localities(local_categories, articles[candidate])
+
+        return recommendations
+
+    def compute_local_dist(self, candidate_articles: ArticleSet):
```
karlhigley, line 125: At first I thought this was computing a distance before figuring out from the return value that it's forming a distribution. Maybe it would be clearer if it was named with preferences in mind?
```suggestion
    def compute_local_prefs(self, candidate_articles: ArticleSet):
```

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -25,13 +47,234 @@ def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestPro
...
+        candidate_articles = candidate_articles.articles
+
+        for article in candidate_articles:
+            candidate_locality = extract_locality(article) or set()
+            for locality in candidate_locality:
+                locality_preferences[locality] += 1
+
+        return locality_preferences
+
+    def compute_topic_dist(self, interest_profile):
```
karlhigley, line 136:
```suggestion
    def compute_topic_prefs(self, interest_profile):
```

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -25,13 +47,234 @@ def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestPro
...
+
+
+model = SentenceTransformer("all-MiniLM-L6-v2")
+
+
+def related_article(current_news_slug, current_topics, clicked_articles, k):
+    # TODO: do we need to consider the cold start?
+    target_embedding = model.encode([current_news_slug])[0]
+
+    subject_candidate_headlines = []
```
karlhigley, line 228: From the code below, I think this is actually a list of subheads? 
```suggestion
    subject_candidate_subheads = []
```

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -25,13 +47,234 @@ def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestPro
...
+        " including the [categories] of news that the user is interested in."
+        " Based on the input news list and user interests,"
+        " you are required to generate a personalized news summary centered around the [main news]."
+    )
+
+    input_prompt = "News List: \n" + f"{news_list}" + "User Interest: \n" + user_interests
+    return gpt_generate(system_prompt, input_prompt)
+
+
+model = SentenceTransformer("all-MiniLM-L6-v2")
```
karlhigley, line 221: Makes sense that you'd need to embed the subheads separately here since the article embedder only does the headlines right now. I'd probably use the same language model for this that the NRMS model is based on, which I think is `distilbert-base-uncased`

Also worth thinking about whether the article embedder should be configurable to specify which article text fields (headline, subhead, body text) to embed. You can do it here but this probably isn't the last time we're going to run into this, so might make sense to move it upstream and make it re-usable.

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -25,13 +47,234 @@ def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestPro
...
             topk=self.num_slots,
         )
-        return ArticleSet(articles=[candidate_articles.articles[int(idx)] for idx in article_indices])
 
-    def add_article_to_categories(self, rec_categories, article):
-        locality_list = extract_locality(article)
-        for locality in locality_list:
-            rec_categories[locality] = rec_categories.get(locality, 0) + 1
+        recommended_list = text_generation(candidate_articles.articles, article_indices, interest_profile)
+        return ArticleSet(articles=recommended_list)
```
karlhigley, line 58: This seems like a whole separate component to me, since the text generation gets applied to every selected article and isn't otherwise coupled to the calibration code. I'd probably use the `article_indices` and `candidate_articles` to build and return an `ArticleSet` here: 
```suggestion
        return ArticleSet(articles=[candidate_articles.articles[idx] for idx in article_indices])
```
and then wire that output into a new component (called something like `ContextGenerator`?) that does the text generation part. More on this in the `text_generation` code below.

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -25,13 +47,234 @@ def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestPro
...
+        subject_candidate_headlines.append(slug)
+
+    candidate_embeddings = model.encode(subject_candidate_headlines)
+    similarities = cosine_similarity([target_embedding], candidate_embeddings)[0]
+    top_k_indices = np.argsort(similarities)[-k:][::-1]
+
+    return top_k_indices
+
+
+def text_generation(candidate_articles, article_indices, interest_profile):
```
karlhigley, line 243: It looks like `candidate_articles` and `article_indices` are used here to reconstruct the equivalent of an `ArticleSet` of the selected articles. If this turns into a separate component, then the `__call__()` method could receive the same information using an `ArticleSet` and an `InterestProfile` as inputs.

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -25,13 +47,234 @@ def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestPro
...
+                [mention.entity.name for mention in clicked_articles[int(idx_clicked)].mentions]
+            )
+            temp_article["topics"] = news_factor(
+                clicked_articles[int(idx_clicked)].headline, clicked_articles[int(idx_clicked)].subhead
+            )
+            news_list.append(temp_article)
+
+        candidate_articles[int(idx)].headline = generate_narrative(news_list, interest_profile.user_interests)
+        recommended.append(candidate_articles)
+    return recommended[0]
```
karlhigley, line 280: I could be missing something, but it seems like this should return more than one article?
```suggestion
    return recommended
```

src/poprox_recommender/handler.py
```diff
@@ -3,8 +3,9 @@
 
 from poprox_concepts import ArticleSet
 from poprox_concepts.api.recommendations import RecommendationRequest, RecommendationResponse
+from poprox_recommender.components.diversifiers.locality_calibration import user_interest_generate
```
karlhigley, line 6: The handler shouldn't know anything about individual recommenders, so importing code from this recommender-specific component is a clue that the code below is probably in the wrong place. Could these changes live somewhere in the `locality_calibration` module?

karlhigley (commented):
I think it makes sense to have a variety of components available in the core library, so I'm generally fine with adding locality-based calibration and context generation here, but as written I think this crosses the line where it makes more sense for this code to live in a fork of the repo. In particular, it's fine if you want to use `openai` in your experiment, but I'm very hesitant to add it as a general dependency of this library that everyone will need to install in order to build recommenders. I'm also a bit concerned about making calls out to an external service—it might work fine but that makes the recommender vulnerable to upstream issues with the OpenAI API that are outside our control. 

I can see two potential paths forward, depending on how important OpenAI is for what you want to do:
* Use a local text generation model from Hugging Face (since we're already depending on `transformers`)
* Move this code to a fork of the repo and invest some time in making the platform more resilient to timeouts and failures

Thoughts on which seems more workable?

mdekstrand:
I agree with @karlhigley's comments on dependencies — let's keep everything in the main recommenders repo self-contained (no external services).

sophiasun0515 (commented):
Looks good to me overall! I think we can determine the number of clicked articles (k) after doing some internal testing. 

For cold start implementation, I don't see the current need for that (assuming we can exclude inactive users in the pre-experiment observation window); 

As for the clicked article sorting, I think it'll be useful to apply it to count for user memory decay (i.e. a clicked article last week would be more salient to users than one from 2 months ago). Maybe we can fuse the chronological factor into embedding similarity computation so there's another layer to consider for candidate article selection other than content embeddings?

src/poprox_recommender/components/diversifiers/locality_calibration.py
```diff
@@ -25,13 +47,234 @@ def __call__(self, candidate_articles: ArticleSet, interest_profile: InterestPro
...
+    for idx in range(len(clicked_articles)):
+        slug = clicked_articles[int(idx)].subhead
+        # category = set([mention.entity.name for mention in clicked_articles[int(idx)].mentions])
+
+        # if any(item in current_topics for item in category):
+        subject_candidate_headlines.append(slug)
+
+    candidate_embeddings = model.encode(subject_candidate_headlines)
+    similarities = cosine_similarity([target_embedding], candidate_embeddings)[0]
+    top_k_indices = np.argsort(similarities)[-k:][::-1]
```
sophiasun0515, line 238: Here's what I'm thinking if we can fuse chronological order into candidate selection

src/poprox_recommender/components/diversifiers/locality_calibration.py
zentavious: This was a challenge I was referring to a few weeks ago over Slack. Text generation shouldn't be applied to every article, so we'd need to set a flag of sorts if we want to pull out this functionality to another component.

karlhigley: Yeah, you can set whatever properties you want on `ArticleSet`, so you could add a list or tensor with boolean flags to indicate whether you want it applied or not and that'll get sent along to subsequent components. I might be missing something but I don't see where in the current code the equivalent selection is happening—it appears to be applying LLM generated text to all selected/recommended articles?

src/poprox_recommender/components/diversifiers/locality_calibration.py
imrecommender: A rough idea—in general, the calculation method for calibration in this function should be similar regardless of the number of dimensions (i.e., preference distribution and defined theta). So, would it be possible to make the preference and theta inputs into a one-to-one matched list (i.e. [topic preference, locality preference], [topic theta, locality theta])? Then we could enumerate the calibration terms in the list🤔

src/poprox_recommender/components/diversifiers/locality_calibration.py
imrecommender: I’m thinking, for handling headlines, should we use the original distilbert transformer, or can we easily extract embeddings from our fine-tuned distilbert on the MIND data? I agree that making this part more flexible for selecting configurations and text fields would be beneficial.

src/poprox_recommender/handler.py
imrecommender: Moved to` locality_calibration `module

src/poprox_recommender/components/diversifiers/locality_calibration.py
imrecommender: Have moved `text_generation` out of the` __call__` method.

src/poprox_recommender/components/diversifiers/locality_calibration.py
karlhigley: It seems like it should be possible to use the fine-tuned one by loading the fine-tuned weights into a plain distilbert model

src/poprox_recommender/handler.py
imrecommender: Just to confirm, could we modify the code in the handler.py in our forked repository?

karlhigley:
Closing this PR since the code has moved to another repository

PR #122, opened 2024-10-30
Save recommendations in DuckDB

mdekstrand:
This PR adds support for saving recommendations in DuckDB instead of Parquet, to facilitate easier storage of additional information with the recommendations (as per Slack discussion) as well as more efficient evaluation (TBD).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,6 +150,12 @@ bower_components/
 Thumbs.db
 [Dd]esktop.ini
 $RECYCLE.BIN/
+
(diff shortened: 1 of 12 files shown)


mdekstrand:
This is currently slow as dirt.

mdekstrand:
This is no longer slow as dirt.

mdekstrand:
I have successfully run this to generate recommendations, but the output files is 10x the size of the Parquet.

PR #123, opened 2024-10-30
Add a components for Reciprocal Rank Fusion

Mahamudul42:
RRF Ranking Code.

### DIFF
--- a/src/poprox_recommender/components/joiners/rrf.py
+++ b/src/poprox_recommender/components/joiners/rrf.py
@@ -11,7 +11,7 @@ class ReciprocalRankFusion(Component):
 
     def __call__(self, candidates1: ArticleSet, candidates2: Lazy[ArticleSet]) -> ArticleSet:
         articles = candidates1.articles
-        article_scores = defaultdict(int)
+        article_scores = defaultdict(float)
         articles_by_id = {}
 
         i = 1
--- a/src/poprox_recommender/components/joiners/rrf.py
+++ b/src/poprox_recommender/components/joiners/rrf.py
@@ -2,40 +2,33 @@ from collections import defaultdict
 
 from poprox_concepts import ArticleSet
 from poprox_recommender.lkpipeline import Component
-from poprox_recommender.lkpipeline.types import Lazy
 
 
 class ReciprocalRankFusion(Component):
-    def __init__(self, num_slots: int):
+    def __init__(self, num_slots: int, k: int = 60):
         self.num_slots = num_slots
+        self.k = k
 
-    def __call__(self, candidates1: ArticleSet, candidates2: Lazy[ArticleSet]) -> ArticleSet:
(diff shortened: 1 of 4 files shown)


src/poprox_recommender/components/joiners/rrf.py
```diff
@@ -0,0 +1,41 @@
...
+from poprox_concepts import ArticleSet
+from poprox_recommender.lkpipeline import Component
+from poprox_recommender.lkpipeline.types import Lazy
+
+
+class ReciprocalRankFusion(Component):
+    def __init__(self, num_slots: int):
+        self.num_slots = num_slots
+
+    def __call__(self, candidates1: ArticleSet, candidates2: Lazy[ArticleSet]) -> ArticleSet:
```
mdekstrand, line 12: Since RRF always uses `candidates2`, it does not need to be `Lazy` (and should not be).

src/poprox_recommender/components/joiners/rrf.py
```diff
@@ -0,0 +1,41 @@
...
+        articles = candidates1.articles
+        article_scores = defaultdict(float)
+        articles_by_id = {}
+
+        i = 1
+        k = 60
+        for article in articles:
```
mdekstrand, line 19: for clarity and error-risk-reduction: this can use `enumerate` rather than manual counters:

```python
for i, article in enumerate(articles, 1):
    ...
```

So can the list below it.

src/poprox_recommender/components/joiners/rrf.py
```diff
@@ -0,0 +1,41 @@
...
+        sorted_article_scores = sorted(article_scores.items(), key=lambda x: x[1], reverse=True)
+
+        sorted_article_ids = [article_id for article_id, article_score in sorted_article_scores]
+
+        reciprocal_rank_fusioned_articles = []
+        for article_id in sorted_article_ids:
+            reciprocal_rank_fusioned_articles.append(articles_by_id[article_id])
+
+        # Return the resulting ArticleSet, limiting the size to num_slots
+        return ArticleSet(articles=reciprocal_rank_fusioned_articles[: self.num_slots])
```
mdekstrand, line 41: The truncation to `self.num_slots` can come before resolving IDs to articles, so that loop doesn't need to run as many times.

src/poprox_recommender/components/joiners/rrf.py
mdekstrand: I was pondering if vectorizing this computation would be worthwhile; it would be faster, but we would need loops to convert article lists to vectorizable form anyway. It would save dictionary lookups in loops, but might not be worth it. Let's go with it as-is for now and assess performance impact later.

mdekstrand (approved):
The logic looks sound. I've left a couple of clarity and performance / straightforwardness comments.

PR #124, opened 2024-11-01
Write a file of embeddings to Parquet from eval recs generation

karlhigley:
This grabs the candidate embeddings from each pipeline execution and builds up a cache of embeddings over the course of the eval run. Once the eval run finishes, it writes them to a two column Parquet file where the first column is the article id and the second is the corresponding embedding.


sophiasun0515 (approved):
lgtm!

mdekstrand (approved):
This looks good to me for doing the saving, but will leave the code in an incompatible state for doing the measurements and automating the run — up to you if you want to fix that in this PR or a second PR.

I like the design of specifying an output directory, this will give us future flexibility to add more output files.

Things that are needed to finish wiring into measurement:

1. `dvc.yaml` — change the `recommend-` stages to specify an output directory, and correct the dependency and output files.
2. update `measure.py` to take an input directory, and look for `recommendations.parquet` in it.
3. `dvc.yaml` — update the `measure-` stages as well.

dvc.yaml
```diff
@@ -1,37 +1,37 @@
 stages:
   recommend-mind-val:
-    cmd: python -m poprox_recommender.evaluation.generate -M MINDlarge_dev -o outputs/mind-val-recommendations.parquet
+    cmd: python -m poprox_recommender.evaluation.generate -M MINDlarge_dev -o outputs/mind-val/
     deps:
       - src/poprox_recommender/evaluation/generate.py
       - data/MINDlarge_dev.zip
     outs:
-      - outputs/mind-val-recommendations.parquet
+      - outputs/mind-val/recommendations.parquet
```
mdekstrand, line 8: We should add `outputs/mind-val/embeddings.parquet` to this as well.

dvc.yaml
```diff
@@ -1,37 +1,37 @@
...
 
   recommend-mind-small:
-    cmd: python -m poprox_recommender.evaluation.generate -M MINDsmall_dev -o outputs/mind-small-recommendations.parquet
+    cmd: python -m poprox_recommender.evaluation.generate -M MINDsmall_dev -o outputs/mind-small/recommendations.parquet
     deps:
       - src/poprox_recommender/evaluation/generate.py
       - data/MINDsmall_dev.zip
     outs:
-      - outputs/mind-small-recommendations.parquet
+      - outputs/mind-small/recommendations.parquet
```
mdekstrand, line 28: Embeddings is also an output here.

dvc.yaml
mdekstrand, line 8: We should add `outputs/mind-val/embeddings.parquet` to this as well.

dvc.yaml
```diff
@@ -1,37 +1,37 @@
...
     deps:
       - src/poprox_recommender/evaluation/evaluate.py
       - data/MINDlarge_dev.zip
-      - outputs/mind-val-recommendations.parquet
+      - outputs/mind-val/recommendations.parquet
     outs:
       - outputs/mind-val-user-metrics.csv.gz
     metrics:
       - outputs/mind-val-metrics.csv:
           cache: false
```
mdekstrand, line 20: Do we want to move the metric output (both declared here and in the `evaluate.py` code) into the run's directory? (do we want to do that as a later PR?)

mdekstrand: “Do we” is a genuine question here — it's fine to merge this and decide we want to move the metric outputs later.

mdekstrand (approved):
Need to add the `embeddings.parquet` files to `outs` in `dvc.yaml` (so DVC will track / push / pull them), but otherwise looks good. One further comment on a now vs. later consistency refactor.

dvc.yaml
karlhigley: That sounds like a good idea

PR #127, opened 2024-11-05
Parallelize generation and re-generate offline evaluation outputs

mdekstrand:
This brings us **parallel generation** of recommendations for offline evaluation, and uses that to regenerate our offline recommendation outputs (and update our reported timings).

This is a big PR, unfortunately, for a couple of reasons:

- Parallelism requires refactoring quite a bit to have the parallel operation encapsulated in a worker function.
- `ipyparallel` doesn't work well with code defined in the script that is being run, so most of the code — in particular, the class definitions it uses, but also functions — are refactored out into imported modules, so the `ipyparallel` workers can find them correctly.

Highlights:

- Parallelize generation with the `ipyparallel` package. I have found this to work well without some of the intermittent bugs I've encountered with multiprocessing or `ProcessPoolExecutor`, and it also makes it easy to run initialization and finalization tasks on all worker threads to set up outputs or finish writing outputs.
- Refactor `generate` into a package (with `__main__.py`, so we can still run it with `python -m poprox_recommender.evaluation.generate`), with various modules implementing its pieces.
- `recommendations` is now a directory instead of a single Parquet file; software that reads Parquets (including Pandas `read_parquet` and DuckDB) generally accept directories of Parquet files and concatenate the files they contain, to support sharding. We use this to shard the output based on worker process; each worker writes its own Parquet output. Otherwise sending data back to the parent and collecting or writing it becomes a bottleneck that severely impairs parallelism (I have learned this through long experience with Lenskit).
- Since each worker has its own output, we can no longer de-duplicate embeddings while we write them. The solution I implemented is that each worker de-duplicates embeddings within its own output, and writes the results to a shared in a temporary directory; the parent process then collects all those embeddings, de-duplicates them once more, and writes the result to a single Parquet file.
- To support the per-worker writing, and to avoid slurping all outputs into memory, this adds a batched parquet writer that accumulates batches of rows and writes them directly to the Parquet file.
- We can still run in a single process with `-j 1` to the evaluation script. The default is to look at the `POPROX_NUM_CPUS` environment variable, or the min of the number of CPUs and 4.
- This adds `recommend-mind-subset` and `measure-mind-subset` tasks to generate and measure for a small subset (first 1K rows) of MIND, to more easily facilitate quick testing of the offline eval code.
- The `poprox_recommender.evaluation.generate.outputs` module defines the layout of the outputs from recommendation generation, so this is defined in one place instead of needing to keep the different pieces of the generator consistent.
- **Renames** the “MRR” user metric to “RR” — the individual per-list metric is recip. rank, and the mean of them is the mean recip. rank.

Work to finish this up:

- [x] parallelize generation
- [ ] update measurement to read from updated output locations
- [ ] update measurement output locations to be consistent with generator outputs
- [ ] rerun everything and collect relevant measurements


karlhigley (approved):

mdekstrand:
@karlhigley FYI this now adds a change to how we detect whether recommendations are “personalized” — it looks at whether the `ranker` produced any outputs, since this is the personalized branch of the pipeline. I don't think this adds any more dependencies on our pipeline structure than are already there; when we look at how to make sure the measurement generalizes beyond our architecture, we should review this.

It speeds up evaluation, because we don't need to look that up in the input data anymore.

PR #128, opened 2024-11-05
Create data loader for exported POPROX data

karlhigley:


mdekstrand:
Over in #127, to reduce overhead in measurement, I added a method `user_has_history(self, user: UUID) -> bool` that just checks if a user has history; the measurement code was constructing entire requests just to see if the user had history (and therefore had personalized recommendations).

However, whether the recommendations are personalized seems like something we should actually be recording at recommendation time, so I'm not super happy with that solution.

mdekstrand:
Actually, never mind that — I can do a cleaner fix in #127.

mdekstrand (approved):
LGTM

sophiasun0515 (approved):
looks good!

src/poprox_recommender/evaluation/evaluate.py
```diff
@@ -25,75 +29,94 @@
 from docopt import docopt
 from progress_api import make_progress
 
+# from poprox_recommender.data.mind import MindData
```
karlhigley, line 32:
```suggestion
```

PR #129, opened 2024-11-06
Locality calibration + text generation

imrecommender:
The main logic in the code includes: 
1) the option to enable or disable text generation
2) the ability to choose whether to consider topic or other filters when finding similar articles from the clicked history
3) the option to apply a time decay weight when identifying similar articles. 

I ran a test on different config combinations using the uploaded basic request JSON. There don’t seem to be any bugs at this point, but the details of the output still need to be checked.


karlhigley:
Closing since the code has moved to another repo

PR #131, opened 2024-11-08
Add a CloudFormation template and CF deploy script

karlhigley:
This sets up the IAM group and user required to deploy the recommender to an outside AWS account. I opted to keep the group (rather than collapse those permissions onto the user) in order to smooth the path for setting up additional users with the same permissions if desired.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -16,8 +16,18 @@ region=${region:-us-east-1}
 echo "ENV: $env"
 echo "Region: $region"
 
-# Download model artifacts
-dvc pull -R models
+# Check if a script name was provided as a command line argument
+if [ -z "$1" ]; then
+    echo "Error: Not sure whether to run cloudformation or serverless"
+    exit 1
+fi
 
-# Build container and deploy functions
-npx serverless deploy --stage "${env}" --region "${region}"
+if [ "$1" == "cf" ]; then
+    aws cloudformation deploy --template-file cloudformation.yml --stack-name poprox-research-recommender-"${env}" --region "${region}" --capabilities CAPABILITY_NAMED_IAM
+elif [ "$1" == "sls" ]; then
+    # Download model artifacts
+    dvc pull -R models
+
+    # Build container and deploy functions
+    npx serverless deploy --stage "${env}" --region "${region}"
+fi
--- a/deploy.sh
+++ b/deploy.sh
@@ -16,6 +16,8 @@ region=${region:-us-east-1}
 echo "ENV: $env"
 echo "Region: $region"
 
+shift $((OPTIND - 1))
+
 # Check if a script name was provided as a command line argument
 if [ -z "$1" ]; then
     echo "Error: Not sure whether to run cloudformation or serverless"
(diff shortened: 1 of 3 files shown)


rburke2233:
It is part of the way there. The problem is that the `dvc pull` step and the `deploy` step need different sets of AWS credentials. So, I ended up writing this, which is indisputably ugly and not proposed as a fix, but works for my setup:

```
elif [ "$1" == "sls" ]; then
	# Set up POPPROX credentials
	source .env
    # Download model artifacts
    dvc pull

	# Set up CU credentials
	source .env.cu
    # Build container and deploy functions
    npx serverless deploy --stage "${env}" --region "${region}"
	
	unset AWS_ACCESS_KEY_ID
	unset AWS_SECRET_ACCESS_KEY
fi
```

deploy.sh
```diff
@@ -16,8 +16,20 @@ region=${region:-us-east-1}
...
+if [ -z "$1" ]; then
+    echo "Error: Not sure whether to run cloudformation or serverless"
+    exit 1
+fi
+
+if [ "$1" == "cf" ]; then
+    aws cloudformation deploy --template-file cloudformation.yml --stack-name poprox-research-recommender-"${env}" --parameter-overrides env="${env}" --region "${region}" --capabilities CAPABILITY_NAMED_IAM
+elif [ "$1" == "sls" ]; then
+    # Download model artifacts
+    dvc pull
```
mdekstrand, line 31: I mad the appropriate change also over in #132, but this can be `dvc pull -R models` to avoid pulling evaluation data that isn't necessary for image deployment.

mdekstrand (approved):
Do I understand this correctly that the result is that we need to do 2 runs:

- `./deploy.sh cf` to set up the CloudFormation skeleton
- `./deploy.sh sls` to deploy the actual endpoints

or does `cf` subsume?

Either way, what I can check looks fine, made a comment on incorporating a small speedup. I would like to see some docs on `deploy.sh` and its usage, where would be the best place to write those? (don't need to block merging on that)

karlhigley:
> Do I understand this correctly that the result is that we need to do 2 runs:
> 
>     ./deploy.sh cf to set up the CloudFormation skeleton
>     ./deploy.sh sls to deploy the actual endpoints
> 
> or does cf subsume?

Yes it currently requires two runs on the first deploy, but the CloudFormation part only needs to be run once and subsequent deploys can just run `./deploy.sh`. If one of them is going to subsume both it should be `sls`. Need to double-check that the CF deploy can be run multiple times without CF stack changes though. 

> Either way, what I can check looks fine, made a comment on incorporating a small speedup. I would like to see some docs on deploy.sh and its usage, where would be the best place to write those? (don't need to block merging on that)

I figure this should probably go in the README, but I'm not in a hurry to merge this PR until @rburke2233 gets a chance to try it out and verify that it works for his scenario.

mdekstrand:
@karlhigley Sounds good. Let's not worry about the script side of subsuming at this point, let's get it tested and — if successful — merged.

We may want to evaluate our use of shell scripts for various tasks and consider moving that kind of logic into a mechanism that is more self-documenting, like tasks for [Invoke](https://www.pyinvoke.org/) (to propose an option that doesn't require anyone to learning a new language).

karlhigley:
> We may want to evaluate our use of shell scripts for various tasks and consider moving that kind of logic into a mechanism that is more self-documenting, like tasks for [Invoke](https://www.pyinvoke.org/) (to propose an option that doesn't require anyone to learning a new language).

Sure! No objection in the slightest, but yet another of a thousand and one nice-to-haves that won't rise to the top of my to-do list any time soon. Would happily merge PRs that make that change in `recommender` and `platform` though.

karlhigley:
This breaks the tests in a way I'm not yet sure how to fix, because running the tests in an offline environment requires a different IAM role configuration than the online setup now uses.

karlhigley:
Closing since it seems like asking for more permissions ended up being a better solution. If we run into this again, we can resurrect this PR, but for now it seems like extra complexity we may not need.

PR #132, opened 2024-11-09
Small Docker tweaks and cleanups

mdekstrand:
Several relatively small improvements to our Docker config:

- configure Serverless to build for `linux/amd64` explicitly
- update deploy.sh to run serverless properly with npx, and only pull necessary DVC data
- bump Pixi version in docker images
- clean up docker image a bit, removing DNF caches (saves 60MB in final image)
- add `outputs/` and `.pixi/` to `.dockerignore`, speeding up Docker build startup a lot

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN pixi install -e pkg
 # Download the punkt NLTK data
 RUN pixi run -e production python -m nltk.downloader -d build/nltk_data punkt
 # Install poprox-recommender
-RUN pixi run -e production pip install --no-deps .
+RUN pixi run -e production pip install --no-deps --root-user-action .
 # Pack up the environment for migration to runtime
 RUN ./.pixi/envs/pkg/bin/conda-pack -p .pixi/envs/production -d /opt/poprox -o build/production-env.tar
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,3 @@
-ARG LOG_LEVEL=INFO
-
 # Use Lambda "Provided" base image for the build container
 FROM public.ecr.aws/lambda/provided:al2023 AS build
 ARG PIXI_VERSION=0.35.0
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -27,17 +25,18 @@ RUN pixi install -e pkg
...
 
 # Use Lambda "Provided" base image for the deployment container
 # We installed Python ourselves
 FROM public.ecr.aws/lambda/provided:al2023
+ARG LOG_LEVEL=INFO
 
 # Unpack the packaged environment from build container into runtime contianer
 # GNU tar chokes on conda-pack's output, so we use bsdtar
-RUN dnf install -y bsdtar
+RUN dnf install -y bsdtar && dnf clean -y all
```
mdekstrand, line 39: This is cleaning up the DNF caches, saving about 60Mb in the final image. I don't do it in the build image since we aren't deploying that image.

serverless.yml
```diff
@@ -28,6 +28,7 @@ provider:
     images:
       poprox-recommender:
         path: "./"
+        platform: linux/amd64
```
mdekstrand, line 31: Explicit platform specification fixes one of the problems Robin faced.

deploy.sh
```diff
@@ -17,7 +17,7 @@ echo "ENV: $env"
 echo "Region: $region"
 
 # Download model artifacts
-dvc pull
+dvc pull -R models
 
 # Build container and deploy functions
-serverless deploy --stage "${env}" --region "${region}"
+npx serverless deploy --stage "${env}" --region "${region}"
```
mdekstrand, line 23: fixing stale script problem (do we want to use `deploy.sh` in the automatic deploys too?)

Dockerfile
```diff
@@ -27,17 +25,18 @@ RUN pixi install -e pkg
 # Download the punkt NLTK data
 RUN pixi run -e production python -m nltk.downloader -d build/nltk_data punkt
 # Install poprox-recommender
-RUN pixi run -e production pip install --no-deps .
+RUN pixi run -e production pip install --no-deps --root-user-action=ignore .
```
mdekstrand, line 28: This just silences a warning that we don't need.

deploy.sh
karlhigley: Yeah, we should do it the same way in our deploy that we expect other people to use

mdekstrand: @karlhigley just pushed a new change that updates the deploy workflow to do this.

karlhigley (approved):

PR #135, opened 2024-11-18
Lazy set of recommendations

karlhigley:
This allows building up a set of recommendations by adding operations that are lazily executed when the final set of recommended items is requested. This accomplishes a few things:
* Moves configuration of the number of slots out of pipeline components
* Enables constraining item positions ("must appear above position 5")
* Allows multiple components to collaborate on building a set of recs


karlhigley:
Closing as superfluous for now, but still interested in finding ways to do some of these things whether that ends up looking like this PR or not.

PR #140, opened 2024-12-03
Sushobhan/update/lenskit

sushobhan2024:
In this PR, we have:

- updated `item_id` and `profile_id` in `evaluation/generate/worker.py`
- updated `evaluation/metrics` to use new LensKit metrics.


src/poprox_recommender/evaluation/metrics/__init__.py
```diff
@@ -35,16 +37,21 @@ def measure_profile_recs(profile: ProfileRecs) -> list[dict[str, Any]]:
     Measure a single user profile's recommendations.  Returns the profile ID and
     a data frame of evaluation metrics.
     """
-    profile_id, all_recs, truth = profile
-    truth.index = truth.index.astype(str)
+    profile_id, all_recs, test = profile
+    test.index = test.index.astype(str)
+
+    filtered_test = test[test["rating"] > 0]
+    test_list = ItemList.from_df(filtered_test)
```
mdekstrand, line 44: If I remember how the test data is indexed correctly, I don't think this will work — I think it has the item IDs on the index, and they need to be in a column.

mdekstrand: May also need to change the test data loader.

sushobhan2024:
Updates:
- decided to stick with `truth` instead of `test` to avoid confusion it may create because we have used `truth` in data loader. 
- renamed index `item` to `item_id` in `profile_truth` in data loader.
- reset index in `metrics` so we have `item_id` as column. however, this has not solved the issue entirely because after running  evaluation or `dvc repro`, we are getting zeros for NDCG and RR. 
Screenshot 2024-12-07 at 5 43 24 AM

sushobhan2024:
Update: 
- Running evaluation now computes NDCG and RR.

src/poprox_recommender/data/mind.py
```diff
@@ -94,8 +94,8 @@ def split_records():
             split_records(),
             columns=["mind_item_id", "rating"],
         )
-        truth["item"] = [self.news_uuid_for_id(aid) for aid in truth["mind_item_id"]]
-        return truth.set_index("item")
+        truth["item_id"] = [self.news_uuid_for_id(aid) for aid in truth["mind_item_id"]]
+        return truth.set_index("item_id")
```
mdekstrand, line 99: I think we need to make the corresponding change in the POPROX loader as well.

mdekstrand (approved):

src/poprox_recommender/data/mind.py
sushobhan2024: I made the change.

PR #143, opened 2024-12-09
added topic-wise-user-embedding

Sris-Ty:
added fundamental static definitions of topics and turned them into virtual clicks according to user preferences and linearly added them with the user embedding

### DIFF
--- a/src/poprox_recommender/components/embedders/topic_wise_user.py
+++ b/src/poprox_recommender/components/embedders/topic_wise_user.py
@@ -1,0 +1,194 @@
+from datetime import datetime, timezone
+from uuid import uuid4
+
+import torch as th
+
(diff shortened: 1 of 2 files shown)


karlhigley:
@mdekstrand There's one test failing here because it uses the model data. I didn't see an existing way to mark tests to be skipped in that context—should we create one?

mdekstrand:
Would love to, but not sure how — the first problem with trying to add a marker for "needs model data" is that there isn't a good way to test "are the models available?" — each individual model is either present or absent. Two paths I see:

1. Add the ability to test for model file existence (either add a `model_file_exists` function, or add an option to `model_file_path` to turn off the existence check), and then use `@pytest.mark.skipif` to skip the test if the model file does not exist.
2. In the test, catch the file-not-found error on the instantiation and call `pytest.xfail` (expected failure).

PR #144, opened 2024-12-30
Address warnings, including attention mask issue that changes the recs

karlhigley:
This addresses three warnings of varying importance:
* The LK pipeline code warned that `SoftmaxSampler` didn't have an annotated return type
* The HF Tokenizers library warned about a tokenization space clean-up behavior that has been enabled by default but will be disabled in future versions
* The HF Transformers library warned about the absence of an attention mask being passed to DistilBERT, which can (and does!) affect the article embeddings

### DIFF
--- a/src/poprox_recommender/components/embedders/article.py
+++ b/src/poprox_recommender/components/embedders/article.py
@@ -53,7 +53,7 @@ class NRMSArticleEmbedder(Component):
         plm_path = model_file_path(config.pretrained_model)
         logger.debug("loading tokenizer from %s", plm_path)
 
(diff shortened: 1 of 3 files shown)


karlhigley:
@sophiasun0515 The change that adds an attention mask to the DistilBERT input changes eventual NRMS output ranking, which makes me wonder if we should retrain the model with the attention mask in place. Thoughts?

mdekstrand:
@karlhigley the test failure should be resolved now with a rerun (I fixed the shared data publication on `main`).

mdekstrand (approved):

sophiasun0515 (approved):
Looks good. Thanks for fixing this!

PR #147, opened 2025-01-07
Add an onboarding topic-aware user embedder with various embedding sources and strategies

Sris-Ty:


karlhigley:
@mdekstrand The action that re-runs the MIND subset recommendations is doing weird things on this PR.

- Is it supposed to run here, or only when we merge code?
- If it is supposed to run here, any idea why it's generating thousands and thousands of log lines loading the same model over and over again? I do see similar log lines when I run it locally, but nearly as many.

mdekstrand:
@karlhigley Yes it's supposed to run here — the goal is to make sure that we don't make changes to pipeline components, data, etc. that leave the repository in a state where the evaluations no longer run.

I don't know yet why it is generating so many spurious log events.

mdekstrand:
"everything fails" error source identified, fixing now.

mdekstrand:
Everything-breaks was caused by prefix-dev/pixi#2960, which our use of 'latest' to lock pixi triggered. I've relocked with latest Pixi.

mdekstrand:
#152 does the Pixi compatibility upgrade on `main`, including fixing Docker.

src/poprox_recommender/components/embedders/topic_wise_user.py
```diff
@@ -92,61 +96,187 @@
...
+            self.embedded_topic_articles = self.article_embedder(ArticleSet(articles=TOPIC_ARTICLES))
+
+        topic_embeddings_by_uuid = {
+            article.article_id: embedding
+            for article, embedding in zip(TOPIC_ARTICLES, self.embedded_topic_articles.embeddings)
+        }
+
+        topic_clicks = virtual_clicks(interest_profile.onboarding_topics, TOPIC_ARTICLES)
+
+        embeddings_from_definitions = self.build_embeddings_from_definitions()
```
mdekstrand, line 143: Here's the problem. This is re-run every call, instead of at setup, which looks unnecessary. Further, `build_embeddings_from_definitions` instantiates a model on every call. The result is that every recommendation request re-instantiates an embedding model.

The fix for the time being is to move either the instantiation, or the entire definition embedding, into `__init__`.

The fix in the future (once component caches exist) is to run the topics through the same embedding component as we use for everything else :).

karlhigley:
@mdekstrand Does the failing "Check data pipeline up-to-date" job here imply that I need to run `dvc repro` and commit the results? If so that's fine for me, but how do people without a pile of GPUs under their desk do that?

karlhigley:
@mdekstrand Thanks, that helps! It is possible to give containers CUDA access but I don't think our dev containers have that yet. I'll give these settings a go using the Pixi environment though.

karlhigley:
I'm running the `dvc repro` process but it's going to take awhile. Will submit a separate PR once that finishes, but want to get this merged so we can start sending ourselves some test newsletters with the new pipelines and getting a feel for how well they work when there's a significant click history to draw on.

mdekstrand (approved):

mdekstrand:
@karlhigley @Sris-Ty do you know what would be making the tests take so long?

karlhigley:
No—it's been stuck there for quite awhile. Not sure if it's something about the pipelines that join multiple recommenders?

mdekstrand:
I cancelled and restarted the job, to see if it was an ephemeral hang, but no. We should stick a timeout on that job, but also figure out what's being slow here.

mdekstrand:
The tests run and pass in reasonable time on my macbook.

Sris-Ty:
In that case, I am hoping joining different recommender is okay as it is taking reasonable time?

mdekstrand:
@Sris-Ty I am not sure what you mean.

I doubt the joining is the problem, as the tests ran fine on my local machine. I am testing in the devcontainer to see if they fail there.

mdekstrand:
glad we could merge this!

PR #149, opened 2025-01-09
Validate output format and edge cases

sushobhan2024:
- [ ] Validate article format
- [ ] Validate minimal set of onboarding topics
- [ ] Validate no clicks and no onboarding cases

### DIFF
--- a/tests/web_service/test_edge_cases.py
+++ b/tests/web_service/test_edge_cases.py
@@ -1,0 +1,60 @@
+"""
+Test edge cases
+"""
+
+import logging
+
+from pydantic import ValidationError
+from pytest import skip
+
+from poprox_concepts.api.recommendations import RecommendationRequest
+from poprox_recommender.config import allow_data_test_failures
+from poprox_recommender.paths import project_root
+
+logger = logging.getLogger(__name__)
+
+
+def test_no_clicks_case():
+    """
+    Test the case where the user has no click history.
+    """
+    test_dir = project_root() / "tests"
+    req_f = test_dir / "request_data" / "request_body.json"
+    if allow_data_test_failures() and not req_f.exists():
+        skip("Request file does not exist")
+
+    try:
+        req = RecommendationRequest.model_validate_json(req_f.read_text())
+        req.interest_profile.click_history = []
+
+        assert isinstance(req.interest_profile.click_history, list)
+        assert len(req.interest_profile.click_history) == 0
+
+        print("No clicks case passed")
+    except ValidationError as e:
+        raise e
+
+    logger.info("No clicks case passed validation.")
+
+
+def test_no_onboarding_case():
+    """
+    Test the case where the user has no onboarding topics.
+    """
+    test_dir = project_root() / "tests"
+    req_f = test_dir / "request_data" / "request_body.json"
+    if allow_data_test_failures() and not req_f.exists():
+        skip("Request file does not exist")
+
+    try:
+        req = RecommendationRequest.model_validate_json(req_f.read_text())
+        req.interest_profile.onboarding_topics = []
+
+        assert isinstance(req.interest_profile.onboarding_topics, list)
+        assert len(req.interest_profile.onboarding_topics) == 0
+
+        print("No onboarding case passed")
+    except ValidationError as e:
+        raise e
+
+    logger.info("No onboarding topics case passed validation.")
--- a/tests/web_service/test_recommendation_format.py
+++ b/tests/web_service/test_recommendation_format.py
@@ -1,0 +1,41 @@
+"""
+Test recommendation format
+"""
+
+import logging
+
+from pydantic import ValidationError
+from pytest import skip
+
+from poprox_concepts import ArticleSet
+from poprox_concepts.api.recommendations import RecommendationRequest
+from poprox_recommender.config import allow_data_test_failures
+from poprox_recommender.paths import project_root
+
+logger = logging.getLogger(__name__)
+
+
+def test_recommendation_format():
+    test_dir = project_root() / "tests"
+    req_f = test_dir / "request_data" / "request_body.json"
+    if allow_data_test_failures() and not req_f.exists():
+        skip("Request file does not exist")
+
+    try:
+        req = RecommendationRequest.model_validate_json(req_f.read_text())
+    except ValidationError as e:
+        raise e
+
+    todays_articles = req.todays_articles
+    try:
+        ArticleSet(articles=todays_articles)
+    except ValidationError as e:
+        raise e
+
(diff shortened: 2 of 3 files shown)


tests/web_service/test_recommendation_format.py
```diff
@@ -0,0 +1,41 @@
...
+
+from poprox_concepts import ArticleSet
+from poprox_concepts.api.recommendations import RecommendationRequest
+from poprox_recommender.config import allow_data_test_failures
+from poprox_recommender.paths import project_root
+
+logger = logging.getLogger(__name__)
+
+
+def test_recommendation_format():
```
karlhigley, line 18: This test looks like it's checking more than just the articles, and validating the format of the whole request—which is good! Could we run this test on all the JSON files in the `request_data` directory?

tests/web_service/test_edge_cases.py
```diff
@@ -0,0 +1,60 @@
...
+
+from pydantic import ValidationError
+from pytest import skip
+
+from poprox_concepts.api.recommendations import RecommendationRequest
+from poprox_recommender.config import allow_data_test_failures
+from poprox_recommender.paths import project_root
+
+logger = logging.getLogger(__name__)
+
```
karlhigley, line 15: For these tests, I think we'll get more mileage out of them if we integrate calls to the recommenders (like you mentioned) so that we can verify that these edge cases don't break the recommendation pipelines. Let's talk about how to do that!

tests/web_service/test_recommendation_format.py
sushobhan2024: `test_recommender_api.py` aims to run on all the JSON files using the server less listener and `RequestGenerator`. I tried running it today but the test has been failing and taking too long to give the error message.

sushobhan2024: @karlhigley, we are now able to send dynamic request data to pipelines with out errors. I will push the modified code later today.

karlhigley:
Looking good! Since this is all test code, I'd probably put it in `tests` instead of in `src` though

sushobhan2024:
@karlhigley @mdekstrand and I were thinking to keep the `testing.py` in `src`. I will be revising this code so we have the helper function to validate response and request data initializer in within `testing.py`, and all the pytest fixtures in `web_service`.

src/poprox_recommender/request_generator.py
```diff
@@ -0,0 +1,83 @@
...
+from uuid import uuid4
+
+from pydantic import ValidationError
+
+from poprox_concepts import AccountInterest, Click, InterestProfile
+from poprox_concepts.api.recommendations import RecommendationRequest
+from poprox_recommender.data.mind import MindData
+
+
+class RequestGenerator:
```
mdekstrand, line 13: should this live in `testing`, or does it make more sense as a separate module?

src/poprox_recommender/request_generator.py
```diff
@@ -0,0 +1,83 @@
...
+            )
+            for i in range(num_clicks)
+        ]
+
+        self.past_articles = [self.mind_data.lookup_article(uuid=click.article_id) for click in self.clicks]
+
+    def add_topics(self, topics: List[str]):
+        self.added_topics = [
+            AccountInterest(
+                account_id=uuid4(),
```
mdekstrand, line 52: these should all be the profile ID, not fresh random UUIDs, shouldn't they?

src/poprox_recommender/testing.py
```diff
@@ -128,3 +129,8 @@ def auto_service() -> Generator[TestService, None, None]:
         # use already-running docker
         port = os.environ.get("POPROX_TEST_PORT", "9000")
         yield DockerTestService(f"http://localhost:{port}/2015-03-31/functions/function/invocations")
+
+
+@fixture(scope="session")
+def mind_data():
+    yield MindData
```
mdekstrand, line 136: should be `yield MindData()` — we want to actually instantiate the MIND data object.

tests/web_service/test_basic_request.py
```diff
@@ -0,0 +1,59 @@
...
+        raise e
+
+
+@mark.docker
+@mark.parametrize("pipeline", PIPELINES)
+def test_basic_request(service, mind_data, pipeline):  # noqa: F811
+    """
+    Initialize request data
+    """
+    request_generator = RequestGenerator(mind_data())
```
mdekstrand, line 35: after the change to the fixture, this will become `mind_data` (without parentheses)

tests/web_service/test_basic_request.py
```diff
@@ -0,0 +1,59 @@
...
+            "Oddities",
+        ]
+    )
+    request_generator.set_num_recs(10)
+    req_body = request_generator.get_request()
+
+    logger.info("sending request")
+    response = service.request(req_body, pipeline)
+    logger.info("response: %s", response.model_dump_json(indent=2))
+    assert response.recommendations
```
mdekstrand, line 53: should we make sure that the request account ID is the key in this dictionary? And that its length is exactly 1? (shouldn't be recommending for users we didn't request)

tests/web_service/test_basic_request.py
mdekstrand: This applies to other tests as well.

tests/web_service/test_basic_request.py
mdekstrand: This applies to other tests as well, probably.

tests/web_service/test_basic_request.py
```diff
@@ -0,0 +1,59 @@
...
+    PIPELINES = recommendation_pipelines().keys()
+except Exception as e:
+    warnings.warn("failed to load models, did you run `dvc pull`?")
+    if allow_data_test_failures():
+        skip("recommendation pipelines unavailable", allow_module_level=True)
+    else:
+        raise e
+
+
+@mark.docker
```
mdekstrand, line 29: Do we need `mark.docker`? What is it for?

mdekstrand (commented):
A few comments, and the big needed change is to move the MindData construction into the fixture, but otherwise I think this is doing what we need! Good work!

.github/workflows/run-tests.yml
```diff
@@ -36,7 +36,10 @@ jobs:
 
       - name: Fetch model data
         run: |
-          uv run dvc pull -r public -R models tests
+          uv run dvc pull -r public -R models tests data
```
mdekstrand, line 39: this needs `-r public` removed.

mdekstrand:
Ok, just the docker tests need to be fixed. Those need the data as well in their `dvc pull`.

PR #150, opened 2025-01-13
Support pipeline configuration files

mdekstrand:
This PR is to work on adding pipeline configuration file support.

In its current form, all it does is add draft configuration file definitions of 2 of our pipelines, to see if we are happy with what they might look like in configuration file form. I've built this around a minimal reuse concept, that supports variable/parameter interpolation and lightweight `extends` mechanism to facilitate reuse. The intended resolution process is:

1. Load and parse the pipeline from TOML into Python data structures.
2. If `meta.extends` is defined, load and parse that TOML file, and do a deep merge of the two parsed pipelines. (This happens recursively, if the referenced pipeline also has `meta.extends`. That allows this pipeline def to override any elements of the parent configuration. Since unused inputs are silently ignored, this should mostly work without defining any key-specific merge strategies, so long as we aren't redefining inputs. Future work can add more nuanced merging if needed.)
3. Walk the data structure, looking for variable references ($...) in any string values. Interpolate them using the variables defined in `definitions` or provided by the code loading the configuration file. This is to allow things like parameterizing the number of recommendation slots in the pipeline configuration (like the current configuration code).
4. Validate the resulting object as a `PipelineConfig`.
5. Construct a pipeline from the validated `PipelineConfig`.

I thought briefly about using Jinja templates instead of rolling our own lightweight extension and variable mechanism. That would also work, and the VS Code Jinja package has support for syntax-highlighting TOML with embedded Jinja, and then we would just inherit Jinja's templating and extension capabilities, but I'm not sure the result would actually be easier to understand. Feedback on this point, or on any elements of what the config files might look like, very welcome!

### DIFF
--- a/pipelines/_common.toml
+++ b/pipelines/_common.toml
@@ -28,27 +28,33 @@ inputs = { article_set = "candidate" }
 code = "poprox_recommender.components.embedders.article:NRMSArticleEmbedder"
 config = { model_path = "nrms-mind/news_encoder.safetensors" }
 inputs = { article_set = "candidate" }
+meta.group = "history-based ranking"
 
 [components.user-embedder]
 code = "poprox_recommender.components.embedders.user:NRMSUserEmbedder"
 config = { max_clicks_per_user = 50, model_path = "nrms-mind/user_encoder.safetensors" }
 inputs = { clicked_articles = "history-embedder", interest_profile = "profile" }
+meta.group = "history-based ranking"
 
 [components.scorer]
 code = "poprox_recommender.components.scorers.article:ArticleScorer"
 inputs = { candidate_articles = "candidate-embedder", interest_profile = "user-embedder" }
+meta.group = "history-based ranking"
 
 [components.ranker]
 code = "poprox_recommender.components.rankers.topk:TopKRanker"
 config = { num_slots = "$N" }
 inputs = { candidate_articles = "scorer", interest_profile = "user-embedder" }
+meta.group = "history-based ranking"
 
 [components.topic-filter]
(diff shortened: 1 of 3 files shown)


pipelines/_common.toml
```diff
@@ -0,0 +1,56 @@
...
+
+[components.candidate-embedder]
+code = "poprox_recommender.components.embedders.article:NRMSArticleEmbedder"
+config = { model_path = "nrms-mind/news_encoder.safetensors" }
+inputs = { article_set = "candidate" }
+
+[components.history-embedder]
+code = "poprox_recommender.components.embedders.article:NRMSArticleEmbedder"
+config = { model_path = "nrms-mind/news_encoder.safetensors" }
+inputs = { article_set = "candidate" }
```
mdekstrand, line 30: An idea: to avoid loading the embedder twice, we could add "reuse" capabilities:

```toml
[components.history-embedder]
reuse = "candidate-embedder"
inputs = { article_set = "candidate" }
```

mdekstrand: This bit will require new pipeline logic, not just a configuration file transform layer, but is entirely feasible.

pipelines/_common.toml
```diff
@@ -0,0 +1,62 @@
...
+meta.group = "history-based ranking"
```
mdekstrand, line 31: This is only to allow an auto-generator of configuration graphs to group components together for better-looking charts.

mdekstrand:
Closing this PR pursuant to Slack discussion on a different direction.

PR #153, opened 2025-01-24
Add retargetable test fixtures and test Docker instead of Serverless Local

mdekstrand:
This makes the test suites configurable in terms of their targets to test.

- The various `_service` fixtures will make a POPROX service available to a test case.
- The `auto_service` fixture looks at the `POPROX_TEST_TARGET` environment variable to see which fixture to load. Supported values are `local` (run handlers directly), `docker-auto` (build and run a docker container), and `docker` (connect to already-running Docker container). Eventually it will be able to be a URL to connect to a deployed Lambda endpoint.
- Use the tests instead of the separate test script to test the docker container. This now tests all pipelines after building the container.
- Remove `serverless-offline` and its bits.

### DIFF
--- a/.devcontainer/setup.sh
+++ b/.devcontainer/setup.sh
@@ -10,4 +10,3 @@ pre-commit install-hooks
 
 # install the development environment
 pixi install -e dev
-pixi run -e dev install-serverless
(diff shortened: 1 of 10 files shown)


mdekstrand:
I've looked at the logs to confirm that the successful docker image test was actually hitting the docker endpoint.

karlhigley (approved):

PR #154, opened 2025-02-03
Model training with dvc.

sushobhan2024:
We have updated preprocessing step. 
I am currently updating `train.py`.


src/poprox_recommender/training/train.py
```diff
@@ -0,0 +1,125 @@
...
+    auc, mrr, ndcg5, ndcg10 = evaluate(args, model, device)
+    logger.info(f"evaluation auc={auc}, mrr={mrr}, ndcg5={ndcg5}, ndcg10={ndcg10}")
+
+
+if __name__ == "__main__":
+    import os
+
+    os.environ["CUDA_DEVICE_ORDER"] = "PCI_BUS_ID"
+    os.environ["CUDA_VISIBLE_DEVICES"] = "0"
+    device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
```
mdekstrand, line 98: We need to remove this hard-coded device configuration, and use the POPROX default device instead.

mdekstrand (commented):
We have a little hard-coded logic that still needs to go away, and need to add the model training stage to `dvc.yaml`.

Otherwise I think testing (and fixing any outstanding bugs), possibly along with resolving any lingering output path issues, are all that's left?

mdekstrand:
So this is running fine on [Screamer](https://codex.lenskit.org/hardware/screamer.html), and should have a retrained model by the weekend.

I updated `pyproject.toml` to use CUDA 12.6 instead of 12.4, so I could get working CUDA packages on Linux AArch64 for testing on another training node.

mdekstrand (approved):
I'm satisfied we didn't break the world in terms of metrics, so when the tests pass I think we're good to merge!

mdekstrand:
We are having space problems on the Docker test…need to fix that before merge, or we can't deploy.

karlhigley (approved):

PR #156, opened 2025-02-05
Split `ArticleSet` into `CandidateSet` and `RecommendationList`

karlhigley:
Differentiating these two types allows us to add attributes and methods to each that are specific to either manipulating pools of items or creating ordered lists of items to be recommended, rather than relying on Python primitives and built-ins to create and manipulate lists of articles.

Depends on https://github.com/CCRI-POPROX/poprox-concepts/pull/39


sophiasun0515 (approved):
Looks good!

mdekstrand (approved):

PR #159, opened 2025-02-14
Add support for compressed request bodies using `gzip`

karlhigley:
We're hitting the upper end of the allowed request size and adding multiple image records per article is pushing us over. I think there are two issues that lead there:
* Our request serialization may not be excluding the raw image metadata attribute/column correctly
* The requests are already large but should be very compressible

This PR attempts to address the second one by checking the `Content-Type` header and decoding/decompressing the request body if the header is set to `application/json+gzip+base64`.

### DIFF
--- a/src/poprox_recommender/handler.py
+++ b/src/poprox_recommender/handler.py
@@ -25,6 +25,10 @@ def generate_recs(event, context):
 
     logger.info(f"Headers: {headers}")
 
+    # base64 encoding is applied to our requests by the AWS stack
+    # and compression is applied in our code, which means that
+    # we have to base64 decode first and decompress second
+    # (contrary to the usual expectation)
     body = base64.b64decode(body) if is_encoded else body
     body = gzip.decompress(body) if is_compressed else body
     logger.info(f"Decoded body: {body}")
--- a/src/poprox_recommender/handler.py
+++ b/src/poprox_recommender/handler.py
@@ -1,4 +1,5 @@
 import base64
+import gzip
 import logging
 import os
 
@@ -15,11 +16,17 @@ logger = logging.getLogger(__name__)
 def generate_recs(event, context):
     logger.info(f"Received event: {event}")
 
-    body = event.get("body", {})
-    is_encoded = event.get("isBase64Encoded", False)
     pipeline_params = event.get("queryStringParameters", {})
+    body = event.get("body", {})
+    headers = event.get("headers", {})
+
+    is_encoded = event.get("isBase64Encoded", False)
+    is_compressed = headers.get("Content-Encoding") == "gzip"
+
+    logger.info(f"Headers: {headers}")
 
     body = base64.b64decode(body) if is_encoded else body
+    body = gzip.decompress(body) if is_compressed else body
     logger.info(f"Decoded body: {body}")
 
     req = RecommendationRequest.model_validate_json(body)
--- a/src/poprox_recommender/testing.py
+++ b/src/poprox_recommender/testing.py
@@ -5,6 +5,8 @@ This lives in the main package so it can be easily imported from any of the
 tests, regardless of their subdirectories.
 """
 
+import base64
+import gzip
 import json
 import logging
 import os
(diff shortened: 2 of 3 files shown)


src/poprox_recommender/handler.py
```diff
@@ -17,9 +18,11 @@ def generate_recs(event, context):
 
     body = event.get("body", {})
     is_encoded = event.get("isBase64Encoded", False)
+    is_compressed = event.get("headers", {}).get("Content-Encoding") == "gzip"
     pipeline_params = event.get("queryStringParameters", {})
 
     body = base64.b64decode(body) if is_encoded else body
+    body = gzip.decompress(body) if is_compressed else body
```
mdekstrand, line 25: I had a double-take on whether these operations are in the correct order because an HTTP body with content-encoding `gzip` should have gzip outside of whatever other encoding happens, but if I understand correctly the base64 encoding is when AWS base64-encodes our HTTP body before delivering it to us, we decode to get the original body and decompress to get the content.

If that is correct, would a comment about what's happening here prevent the next reader from being confused as well?

mdekstrand (approved):

src/poprox_recommender/handler.py
karlhigley: That's correct, and I'm happy to add a comment.

mdekstrand:
> @mdekstrand See if this is less double-take-worthy?

If anything I think it's a little more confusing, especially with the bespoke MIME type. Will that MIME type be sent on the lambda path?

My understanding of what is happening with encodings is this:

- We send JSON to the AWS Lambda API gateway. This PR changes so we can send compressed JSON to the API gateway.
- The AWS Lambda API Gateway Base64-encodes whatever payload it receives, stuffs it into the `body` field of a request, and sends it to the actual lambda implementation. I'm not sure why our code needs to un-Base64 it (shouldn't that kind of thing be transparent?), but we do, so oh well.
- Our test code is directly hitting the lambda, not the API gateway, so we need to mimic the Base64 encoding (in addition to the new compression logic).

If this understanding is correct, then Content-Encoding does seem like the “correct” approach (in the sense of using the standards as they are intended, not neceessarily in other senses): we are sending JSON to the REST endpoint (served by API gateway), and we are encoding it with gzip to save space, and the Base64 business is just backend things and a hack around not having the API gateway in our test harness.

If our client code will need to send the `application/json+gzip+base64` content type to make this server logic work, then its content type will actually be incorrect from the perspective of the API gateway endpoint: we'll be sending GZIP-encoded JSON, but calling it json + gzip + base64.

At the end of the day, the most important thing is to have it working. But in the interest of clarity (as well as scrutability to things like HTTP debugging tools), if it works with content-encoding, that seems like a better solution to me.

karlhigley:
I'll give it a third try 😅

karlhigley:
Going with the original version with minor tweaks, which should be backwards compatible since the existing requests aren't sent with the `Content-Encoding`: `gzip` header. I'll test it out by sending recommendation requests for test newsletters from the platform side

karlhigley:
@mdekstrand This sets the headers in what I think is the expected way (`{"Content-Type": "application/json", "Content-Encoding": "gzip"}`) but fails the integration tests with the Docker container with the cryptic error:
```
Traceback (most recent call last):able to unmarshal input: Expecting value: line 1 column 1 (char 0)
```
(and no actual traceback)

I'm not sure how to troubleshoot that since the test passes with the local test service and I'm not getting much in the logs

mdekstrand:
@karlhigley that's confusing and unhelpful lambda behavior :(.

my first attempt would be to move `generate_recs` to `generate_recs_impl` and make `generate_recs` call that in a try/catch, manually print the stack trace before re-throwing the exception. see if we can get a real error message that way.

mdekstrand:
> @karlhigley that's confusing and unhelpful lambda behavior :(.
> 
> my first attempt would be to move `generate_recs` to `generate_recs_impl` and make `generate_recs` call that in a try/catch, manually print the stack trace before re-throwing the exception. see if we can get a real error message that way.

Or maybe that would be cleaner with `generate_recs_wrapper` that does the exception printing.

karlhigley:
@mdekstrand Looks like the secret was to use the `base64.encodebytes()` like the `test-docker-lambda.py` script does

mdekstrand:
that's… obvious.

PR #164, opened 2025-02-27
Update LensKit dependency

mdekstrand:
This updates LensKit to a newer beta, and makes sure everything still works.

**Note:** this changes the default RBP patience parameter from 0.5 to 0.85.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ readme = "README.md"
 license = { file = "LICENSE.md" }
 dynamic = ["version"]
 dependencies = [
-  "lenskit==2025.0.0a4",
+  "lenskit==2025.1.1b12",
   "nltk>=3.8,<4",
   "numpy>=1.26,<2",
   "torch~=2.2",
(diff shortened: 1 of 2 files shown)


src/poprox_recommender/components/embedders/article.py
```diff
@@ -20,6 +21,12 @@
 TITLE_LENGTH_LIMIT = 30
 
 
+@dataclass
+class NRMSArticleEmbedderConfig:
+    model_path: PathLike
+    device: str | None
```
mdekstrand, line 27: This (and the user config) are dataclasses, instead of Pydantic models, because names starting with `model_` are reserved in Pydantic models, but are not reserved in the Pydantic dataclass machinery. Pydantic and LensKit support both `BaseModel` and dataclass configuration classes.

We may want to use dataclass for all component configurations for consistency.

mdekstrand:
ok this is all ready I think.

karlhigley (approved):

PR #165, opened 2025-03-03
Use case-insensitive `Headers` implementation from `httpx`

karlhigley:
It seems that API Gateway lower-cases all the headers in between the platform and the recommender, but out of an abundance of caution let's just make the headers case-insensitive. We already have `httpx` as a transitive dependency, so this doesn't really add any additional heft to the container.


mdekstrand (approved):

karlhigley:
Closing in favor of #166

PR #167, opened 2025-03-03
Add `/warmup` route returning protocol version and available pipelines

karlhigley:
This route accomplishes several things:
* Get recommendation pipelines and models loaded into memory in at least one Lambda instance before we send recommendation requests
* Tell the platform what version of the request/response format this endpoint supports
* Provide the list of available recommenders back to the platform, which allows us to throw better exceptions on the platform side before we ask for a recommender that doesn't exist in the endpoint

It's a bit tricky to test with the current version of the test service, which (reasonably) assumed only one route per endpoint.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,8 +28,7 @@ dependencies = [
   "transformers>=4.41,<5",
   "rich ~=13.9",
   "ipyparallel ~=8.0",
-  # we reference the zip file, so install doesn't need git
-  "poprox-concepts @ https://github.com/CCRI-POPROX/poprox-concepts/archive/refs/heads/main.zip",
+  "poprox-concepts @ git+https://github.com/CCRI-POPROX/poprox-concepts.git",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
   "rich ~=13.9",
   "ipyparallel ~=8.0",
   # we reference the zip file, so install doesn't need git
-  "poprox-concepts @ https://github.com/CCRI-POPROX/poprox-concepts/archive/refs/tags/v0.1.0.zip",
+  "poprox-concepts @ https://github.com/CCRI-POPROX/poprox-concepts/archive/refs/heads/main.zip",
```
mdekstrand, line 32: This is probably a problem — as soon as someone commits to `-concepts`, everyone's lock files on `-recommender` will break. It is annoying but the best practice would be to switch to the Git URLs (`git+https://github.com/CCRI-POPROX/poprox-concepts.git`) when using an in-development version, so `uv` can fetch the precise version referenced in the lock file. Maybe have both versions here with one commented out, and a comment providing instructions to switch?

mdekstrand (approved):

pyproject.toml
karlhigley: Switching to the `git` URL makes sense to me, but not sure why we'd keep both around. It just means that the install will need git.

PR #168, opened 2025-03-05
Update the recommender to use version 2 of the request/response format

karlhigley:
This allows some further clean up of the main route, but more importantly `RecommendationResponseV2` uses `RecommendatList` as the type for the recs, which has the `extras` field that will enable additional custom logging.


sophiasun0515 (approved):
Looks good to me!

PR #169, opened 2025-03-11
Refactor pipelines into separate configurations + cache

mdekstrand:
This is the Pipeline Config Refactor, giving us configuration "files" and caching.

- Move each pipeline configuration into a separate file (with a `common` file for standard elements of the configuration).
- Rewrite the pipeline load logic to scan the avialable files, instead of using a hard-coded list, and load pipelines from there.
- Allow only the desired pipeline to be loaded, instead of loading all.
- Use a `PipelineCache` to automatically share common componens across all loaded pipelines.
- Remove the concept of multiple pipeline names — give all pipelines one name, the name of their module. Right now those names are unchanged from original, I would like to rename for consistency and clarity.

With pipelines in separate files, we will be able to add (in another PR, probaly) logic to generate and measure separately, so we only need to re-run evaluation for the pipeline we add or change instead of re-running _everything_.


mdekstrand:
Once we are happy with this code, I will cut a LensKit release and make us depend on the release again before merging, I just want to test the builder and cache logic in POPROX before cutting that release.

src/poprox_recommender/recommenders/configurations/mmr.py
```diff
@@ -0,0 +1,26 @@
...
+from lenskit.pipeline import PipelineBuilder
+
+from poprox_recommender.components.diversifiers import (
+    # LocalityCalibrator,
+    MMRDiversifier,
+)
+from poprox_recommender.components.embedders import NRMSArticleEmbedder, NRMSUserEmbedder
+from poprox_recommender.paths import model_file_path
+
+from .. import common
```
karlhigley, line 12: I get that there are repeated chunks of code across the pipelines, but this `common` thing seems a little weird to me. It obscures the structure of the pipeline by making the reader bounce between files to figure out what's connected to what. Reversing how `builder` is used from the normal `builder.do_whatever()` style is kinda confusing too.

I'd probably just go ahead and duplicate the code across files for now, figuring that's an acceptable intermediate step between where we were and loading a directory of serialized pipeline definition files

mdekstrand: Yeah, if we do want to abstract a `PoproxPipelineBuilder` class that wraps `PipelineBuilder` is probably a better plan. I'll duplicate the code for now.

mdekstrand:
Still need the release instead of git dep. In flight.

mdekstrand:
Ok, now we have a tagged release to depend on.

karlhigley (approved):

PR #170, opened 2025-03-11
personalized image of users based on caption added here

Mahamudul42:
I’ve completed the coding for selecting the best image based on user preferences. The approach uses similarity between user embeddings and image caption embeddings to personalize image selection.
New files added:
image_selector.py – Contains the select_image function for choosing the best image [src/poprox_recommender/components/filters/image_selector.py].
caption_embedder.py – Handles embedding generation for image captions [src/poprox_recommender/components/embedders/caption_embedder.py].
image-request.json – Stores image request data [tests/request_data/image-request.json].
Modified file:
recommenders.py – Updated the select_articles function to incorporate image selection based on personalized user preferences. Added:
process_pipeline_state function to refine recommended articles with image selection.
RecommendationState class to mimic PipelineState for writable defaults and metadata.
We can run the code using:
python scripts/send-request.py -p 9000 -r tests/request_data/image-request.json

Thanks!

### DIFF
--- a/src/poprox_recommender/components/embedders/caption_embedder.py
+++ b/src/poprox_recommender/components/embedders/caption_embedder.py
@@ -1,0 +1,83 @@
+import logging
+from dataclasses import dataclass
+from os import PathLike
+from typing import List
+
+import torch as th
+from safetensors.torch import load_file
+from transformers import AutoTokenizer
+
+from poprox_recommender.model import ModelConfig
+from poprox_recommender.model.nrms.news_encoder import NewsEncoder
+from poprox_recommender.paths import model_file_path
+from poprox_recommender.pytorch.decorators import torch_inference
+
+logger = logging.getLogger(__name__)
+CAPTION_LENGTH_LIMIT = 30
+
+
+@dataclass
+class CaptionEmbedderConfig:
+    model_path: PathLike
+    device: str | None
+
+
+class CaptionEmbedder:
+    config: CaptionEmbedderConfig
+    model: NewsEncoder
+    tokenizer: AutoTokenizer
+    embedding_cache: dict[str, th.Tensor]
+
+    def __init__(self, config: CaptionEmbedderConfig | None = None, **kwargs):
+        if config is None:
+            config = CaptionEmbedderConfig(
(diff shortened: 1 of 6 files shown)


src/poprox_recommender/components/embedders/caption_embedder.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def __init__(self, config: CaptionEmbedderConfig | None = None, **kwargs):
+        if config is None:
+            config = CaptionEmbedderConfig(
+                model_path=model_file_path("nrms-mind/news_encoder.safetensors"), device="cpu"
+            )
+
+        self.config = config
+
+        checkpoint = load_file(self.config.model_path)
```
karlhigley, line 39: Loading the model here manually would work, but there's a better way: you can use the `NRMSArticleEmbedder` component directly, which knows how to load the model and will be cached to avoid loading it more than once.

Here's an example of what that looks like: https://github.com/CCRI-POPROX/poprox-recommender/blob/main/src/poprox_recommender/components/embedders/topic_wise_user.py#L156-L158

src/poprox_recommender/recommenders.py
```diff
@@ -58,7 +82,45 @@ def select_articles(
     else:
         wanted = (topk, recs)
 
-    return pipeline.run_all(*wanted, candidate=candidate_articles, clicked=clicked_articles, profile=interest_profile)
+    pipeline_state = pipeline.run_all(
```
karlhigley, line 85: We want `pipeline.run_all()` to execute your components too, which means your components need to be added to the pipeline. Here's an example of how pipelines are built: https://github.com/CCRI-POPROX/poprox-recommender/blob/main/src/poprox_recommender/recommenders/configurations/mmr.py

When you call `builder.add_component()`, you provide information about what class to use and how to wire up the inputs to other components in the pipeline. The names of the arguments should match the names on your `__call__()` methods.

karlhigley (commented):
This looks like it should work, so according to the maxim "Make it work, make it right, make it fast" we can move on to making it right now that you've made it work. Here that will mean aligning this code with other code in the repo, which will simplify some of the things you're doing.

PR #172, opened 2025-03-14
Fix default recommender pipeline when pipeline=None

mdekstrand:
When the `pipeline` option is passed to `select_articles` but is expicitly `None`, this was not correctly detected in the default pipeline fallback logic.

This PR corrects that, falling back when either `pipeline` is missing or is `None`. It also adds a configuration for the default pipeline, both in a constant and in an environment variable, with a `default_pipeline` function to look up the default.

### DIFF
--- a/src/poprox_recommender/recommenders/__init__.py
+++ b/src/poprox_recommender/recommenders/__init__.py
@@ -5,8 +5,14 @@ from lenskit.pipeline import PipelineState
 
 from poprox_concepts import CandidateSet, InterestProfile
 
-from .configurations import DEFAULT_PIPELINE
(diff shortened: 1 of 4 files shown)


src/poprox_recommender/recommenders/__init__.py
```diff
@@ -29,10 +36,12 @@ def select_articles(
...
     pipeline state whose ``default`` is the final list of recommendations.
     """
+    name = None
     if pipeline_params and "pipeline" in pipeline_params:
         name = pipeline_params["pipeline"]
-    else:
-        name = "nrms"
+
+    if name is None:
+        name = default_pipeline()
```
mdekstrand, line 44: This is the key part that fixes the bug. The rest is a test + better logic for defining what the default pipeline is.

karlhigley (approved):

PR #175, opened 2025-03-25
Add WRRF in joiners to Support Weighted Fusion of Multiple Recommendation Lists

Mahamudul42:
### Overview
This PR introduces `wrrf.py` to the `poprox_recommender/joiners` directory, implementing Weighted Reciprocal Rank Fusion (WRRF). WRRF extends the existing `rrf.py` by adding support for weighted fusion of recommendation lists, allowing prioritization of more reliable or important sources, and improving the quality of the fused list. The implementation is backward compatible with the existing RRF approach and addresses its key limitations.

### Motivation
The current RRF implementation in `rrf.py` has two main constraints:
1. **Equal Weighting**: It treats all input lists equally, which can dilute the impact of stronger recommendation models when weaker ones are included.
2. **Fixed Two-List Input**: It is limited to fusing exactly two recommendation lists, reducing flexibility for more complex scenarios.

WRRF addresses these issues by:
- Introducing weights (e.g., `{"recs1": 1.5, "recs2": 1.0}`) to emphasize more reliable or important recommendation sources.
- Supporting a flexible number of input lists (minimum of two, with additional lists optional), making it more adaptable to diverse use cases.
- Maintaining backward compatibility by defaulting to equal weighting if no weights are specified.

### Changes
- Added `wrrf.py` to `poprox_recommender/joiners/`, implementing the WRRF algorithm.
- The implementation extends the RRF logic to handle weighted fusion and multiple input lists.

### DIFF
--- a/src/poprox_recommender/components/joiners/wrrf.py
+++ b/src/poprox_recommender/components/joiners/wrrf.py
@@ -1,0 +1,74 @@
+from collections import defaultdict
+from itertools import zip_longest
+from typing import List, Optional
+
+from lenskit.pipeline import Component
+from pydantic import BaseModel
+
+from poprox_concepts.domain import RecommendationList
+
+
+class WRRFConfig(BaseModel):
+    num_slots: int
+    k: int = 60
+    # Default weights if not specified
+    default_weights: dict = {"recs1": 1.0, "recs2": 1.0}
+
+
+class WeightedReciprocalRankFusion(Component):
+    config: WRRFConfig
+
+    def __call__(
+        self,
+        recs1: RecommendationList,
+        recs2: RecommendationList,
+        *additional_recs: Optional[RecommendationList],
+        weights: Optional[dict] = None,
+    ) -> RecommendationList:
+        """
+        Perform Weighted Reciprocal Rank Fusion on multiple recommendation lists.
+
+        Args:
+            recs1: First recommendation list (required)
+            recs2: Second recommendation list (required)
+            *additional_recs: Optional additional recommendation lists
+            weights: Optional dictionary of weights for each list (e.g., {"recs1": 1.5, "recs2": 1.0})
+
+        Returns:
+            RecommendationList: Fused recommendation list
+        """
+        # Combine all recommendation lists into a single list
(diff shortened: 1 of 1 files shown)


karlhigley:
If this is backward compatible then it seems like this could go in the existing `rrf.py` without breaking anything, right?

Mahamudul42:
Yes, since WRRF is backward compatible—defaulting to equal weighting like RRF—it could technically be added to the existing `rrf.py` without breaking anything. However, I think keeping it in a separate `wrrf.py` file might be better for clarity and modularity, as it signals the added weighting functionality as a distinct feature. This way, users can easily choose between standard RRF and weighted fusion based on their needs.

src/poprox_recommender/components/joiners/wrrf.py
```diff
@@ -0,0 +1,77 @@
...
+
+class WeightedReciprocalRankFusion(Component):
+    config: WRRFConfig
+
+    def __call__(
+        self,
+        *,
+        recs1: RecommendationList,
+        recs2: RecommendationList,
+        additional_recs: Optional[List[RecommendationList]] = None,
```
mdekstrand, line 26: Actually using this is slightly difficult (we don't have components that put together multiple recommendation lists into a list).

The best solution to handle more than 2 lists, IMO, is to make it a variadic keyword argument, and probably not have the `recs1` and `recs2` special cases — just take a `**recommendations: RecommendationList`, and then it will work on whatever inputs are wired to it. However, supporting this will also need a change in the pipeline runner: right now the pipeline does not handle variadic components. Adding support for them would not be incredibly difficult, but is a little fiddly.

We can add a 2-list version right now without needing any pipeline changes.

src/poprox_recommender/components/joiners/wrrf.py
```diff
@@ -0,0 +1,77 @@
...
+        weights: Optional[dict] = None,
```
mdekstrand, line 27: Do we envision a component computing weights such that we want this to be a runtime input, rather than just supporting it as a configuration option?

PR #177, opened 2025-03-26
Incorporate padding into attention mechanisms in NRMS

karlhigley:
After this change, we ignore padding in three places in the news encoder:
* At the input to the pre-trained language model since headlines are different lengths (already implemented)
* At the input to the multi-head attention module since headlines are different lengths (available out of the box, applied here)
* At the input to the additive attention module since MHA's outputs are non-zero even for padded/masked positions (newly implemented and applied here)

And in two places in the user encoder:
* At the input to the multi-head attention module since click histories are different lengths (available out of the box, applied here)
* At the input to the additive attention module since MHA's outputs are non-zero even for padded/masked positions (newly implemented and applied here)


mdekstrand (approved):

sophiasun0515 (approved):
LGTM! :shipit:

PR #179, opened 2025-04-02
Addressed feedback from PR #170, also compatible with latest pipeline…

Mahamudul42:
Updated with the latest changes in the pipeline.

### DIFF
--- a/src/poprox_recommender/components/embedders/caption_embedder.py
+++ b/src/poprox_recommender/components/embedders/caption_embedder.py
@@ -1,0 +1,47 @@
+import logging
+from dataclasses import dataclass
+from typing import Sequence
+
+import torch as th
+from transformers import AutoModel, AutoTokenizer
+
+from poprox_recommender.components.embedders.article import (
+    NRMSArticleEmbedder,
+    NRMSArticleEmbedderConfig,  # Added import
+)
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class CaptionEmbedderConfig:
+    model_path: str
+    device: str
+
+
+class CaptionEmbedder:
+    def __init__(self, config: CaptionEmbedderConfig):
+        self.config = config
+
+        # Convert CaptionEmbedderConfig to NRMSArticleEmbedderConfig
+        article_config = NRMSArticleEmbedderConfig(
+            model_path=config.model_path,
+            device=config.device,
+        )
+        self.article_embedder = NRMSArticleEmbedder(config=article_config)
+
+        self.tokenizer = AutoTokenizer.from_pretrained("distilbert-base-uncased")
+        self.encoder = AutoModel.from_pretrained("distilbert-base-uncased").to(self.config.device)
+
+    def __call__(self, captions: Sequence[str]) -> th.Tensor:
+        tokens = self.tokenizer(
+            captions,
+            padding=True,
+            truncation=True,
+            return_tensors="pt",
+        ).to(self.config.device)
+
+        output = self.encoder(**tokens)
+        caption_embeds = output.last_hidden_state[:, 0, :]
+
+        return caption_embeds
--- a/src/poprox_recommender/components/filters/image_selector.py
+++ b/src/poprox_recommender/components/filters/image_selector.py
@@ -1,0 +1,66 @@
+import logging
+from typing import Optional
+
+import torch as th
+
+from poprox_concepts.domain import Image, InterestProfile, RecommendationList
+from poprox_recommender.components.embedders.caption_embedder import CaptionEmbedder
+from poprox_recommender.pytorch.decorators import torch_inference
+
+logger = logging.getLogger(__name__)
+
+
+class ImageSelector:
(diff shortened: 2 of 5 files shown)


src/poprox_recommender/recommenders/configurations/nrms.py
```diff
@@ -44,11 +45,23 @@ def configure(builder: PipelineBuilder, num_slots: int, device: str):
...
+
     # Fallback: sample from user topic interests
     n_topic_filter = builder.add_component(
         "topic-filter", TopicFilter, candidate=i_candidates, interest_profile=i_profile
     )
     n_sampler = builder.add_component("sampler", UniformSampler, candidates1=n_topic_filter, candidates2=i_candidates)
 
     # Combine primary ranker and fallback
-    builder.add_component("recommender", FillRecs, {"num_slots": num_slots}, recs1=n_ranker, recs2=n_sampler)
+    builder.add_component("recommender", FillRecs, {"num_slots": num_slots}, recs1=o_with_images, recs2=n_sampler)
```
karlhigley, line 67: I'd probably put the image selector after the fill to make sure all images are personalized regardless of where the selected items come from:
```suggestion
    n_fill = builder.add_component("fill", FillRecs, {"num_slots": num_slots}, recs1=n_ranker, recs2=n_sampler)
    n_images = builder.add_component(
        "recommender",
        image_selector,
        recommendations=n_fill,
        interest_profile=e_user,
    )
```
By convention, the last node in the pipeline is called `recommender`, which helps us find the right output from among several possibilities.

src/poprox_recommender/components/filters/image_selector.py
```diff
@@ -0,0 +1,66 @@
...
+
+        for article in recommendations.articles:
+            selected_image = self._select_article_image(article, user_embedding)
+            if selected_image:
+                article.preview_image_id = selected_image.image_id
+                # logger.debug(f"Selected image {selected_image.image_id} for article {article.article_id}")
+
+        return recommendations
+
+    def _select_article_image(self, article, user_embedding) -> Optional[Image]:
```
karlhigley, line 40: Adding type annotations would make this a bit easier to read
```suggestion
    def _select_article_image(self, article: Article, user_embedding: th.Tensor) -> Image | None:
```
The `Optional[Image]` type is correct but you can do it without importing from `typing` with the `A | B` syntax

src/poprox_recommender/components/filters/image_selector.py
```diff
@@ -0,0 +1,66 @@
+import logging
+from typing import Optional
+
+import torch as th
+
+from poprox_concepts.domain import Image, InterestProfile, RecommendationList
+from poprox_recommender.components.embedders.caption_embedder import CaptionEmbedder
+from poprox_recommender.pytorch.decorators import torch_inference
+
+logger = logging.getLogger(__name__)
```
karlhigley, line 10: I'd suggest either uncommenting the debug logging below and set the log level on this logger to info or warning to hide them, or removing the commented log statements.

src/poprox_recommender/components/embedders/caption_embedder.py
```diff
@@ -0,0 +1,47 @@
...
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class CaptionEmbedderConfig:
+    model_path: str
+    device: str
+
+
+class CaptionEmbedder:
```
karlhigley, line 22: As currently written, this isn't a component you could wire into a pipeline directly and is only usable together with `ImageSelector`. I'd probably condense this code into `ImageSelector` for now—if we come up with another use for caption embeddings we can split it out into a separate component later, but for now this can all be one component.

src/poprox_recommender/recommenders/__init__.py
```diff
@@ -33,23 +34,30 @@ def select_articles(
...
 
     pipeline = get_pipeline(name)
 
     recs = pipeline.node("recommender")
+    image_selector = pipeline.node("image-selector")
     topk = pipeline.node("ranker", missing="none")
+
     if topk is None:
-        wanted = (recs,)
+        wanted = (image_selector, recs)
```
karlhigley, line 54: By making the output of the image selector the output of the whole pipeline, we'll get the articles with personalized images in the `recs`. This file shouldn't need changes to make your recommender pipeline work.

karlhigley (commented):
This is looking pretty good overall, and could still use a few changes to get it ready to merge.

karlhigley:
The test failures are due to changes in `recommenders/__init__.py`, which make every pipeline try to fetch results from the image selector (even the ones that don't have it.) The suggestions above should resolve that issue.

Mahamudul42:
Update NRMS pipeline to move image selection after FillRecs and combine CaptionEmbedder

- Moved image selection to occur after combining ranked and sampled recommendations
- Combined CaptionEmbedder logic into ImageSelector per review feedback
- Removed commented logging statements
- Ensured final pipeline node is named 'recommender'

karlhigley:
Superseded by #180

PR #182, opened 2025-04-10
Aishwarya/feature/mind evaluation metrics

aisa6148:
Diversity Metrics + Testcases:

1. RBO - fixes and testcases added
2. Gini
3. K-coverage
4. LIP
5. Gini
6. Gini for topics
7. RBE

[Review Sheet](https://docs.google.com/spreadsheets/d/1D_oHFj640T2lZFJgHpDZQ66Kla7yEOQrrSXsd8ic6eY/edit?gid=0#gid=0) -- for calculation reference

### DIFF
--- a/src/poprox_recommender/evaluation/metrics/lip.py
+++ b/src/poprox_recommender/evaluation/metrics/lip.py
@@ -6,7 +6,6 @@ from poprox_concepts.domain import CandidateSet
 def least_item_promoted(reference_article_set: CandidateSet, reranked_article_set: CandidateSet, k: int = 10) -> float:
     if not reference_article_set.articles:
         return np.nan
-
     lip_rank = k
     for item in reranked_article_set.articles[:k]:
         try:


karlhigley:
@mdekstrand It looks like the tests are failing here due to a missing `ray` import. Possible that we still need to add the eval dependencies to the test environment?

src/poprox_recommender/evaluation/metrics/gini.py
```diff
@@ -0,0 +1,17 @@
+from collections import Counter
+
+import numpy as np
+from lenskit.stats import gini
+
+from poprox_concepts.domain import CandidateSet
+
+
+def gini_coeff(final_recs: CandidateSet) -> float:
```
mdekstrand, line 9: it looks like this is trying to compute the gini from a single recommendation list — that should always be 0, shouldn't it?

mdekstrand:
@karlhigley that would work as a stopgap, yes.

mdekstrand:
Metric code restructure would also be my preferred solution. I think it's mostly a matter of pulling some worker code out of `evaluation/__init__.py` into a submodule of `evaluation`.

mdekstrand:
Commit [ba3aeda](https://github.com/CCRI-POPROX/poprox-recommender/pull/189/commits/ba3aedaab0cc91aae6052ce18581cb95398b0b12) in #189 has the fix for this.

src/poprox_recommender/evaluation/metrics/rbe.py
```diff
@@ -0,0 +1,33 @@
...
+
+mind_data = MindData()
+
+
+def rank_bias_entropy(final_recs: CandidateSet, k: int, d: float = 0.5):
+    top_k_articles = final_recs.articles[:k]
+    weighted_counts = defaultdict(float)
+
+    for rank, article in enumerate(top_k_articles):
+        weight = d ** (rank + 1)
```
mdekstrand, line 16: should be `d ** rank` — we want the first rank to have weight 0 (for the RBP geometric cascade).

They aren't in the public docs yet but LensKit now has weighting functions abstracted from the metrics, so we could call those to get weights.

src/poprox_recommender/evaluation/metrics/rbo.py
```diff
@@ -11,14 +11,23 @@ def rank_biased_overlap(recs_list_a: CandidateSet, recs_list_b: CandidateSet, p:
...
+
+    sum = 0
+    weights = 0
     for d in range(1, k + 1):
         set_a = set([a.article_id for a in recs_list_a.articles[:d]])
         set_b = set([b.article_id for b in recs_list_b.articles[:d]])
         overlap = len(set_a.intersection(set_b))
         agreement = overlap / d
-        summands.append(agreement * np.power(p, d))
+        weight = np.power(p, d)
```
mdekstrand, line 27: If we're using the same RBP cascade, this should be `d-1` (as per comment above). Or we should use the abstracted weighting logic.

src/poprox_recommender/evaluation/metrics/rbe.py
```diff
@@ -0,0 +1,33 @@
...
+
+import numpy as np
+
+from poprox_concepts.domain import CandidateSet
+from poprox_recommender.data.mind import MindData  # Import MindData to access lookup_article
+
+mind_data = MindData()
+
+
+def rank_bias_entropy(final_recs: CandidateSet, k: int, d: float = 0.5):
```
mdekstrand, line 11: This default for `d` does not match RBP's default.

tests/test_gini_topics.py
```diff
@@ -0,0 +1,74 @@
+import logging
+import unittest
```
mdekstrand, line 2: We're using PyTest in the rest of the tests — hate to ask for a test rewrite, but can we be consistent on our testing framework?

tests/test_rbe.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def test_entropy_with_same_mentions_present(self):
+        ranked = CandidateSet(articles=[self.m2, self.m4])
+        score = rank_bias_entropy(ranked, k=2)
+        self.assertEqual(score, 0.0)
+
+    def test_entropy_with_diverse_mentions(self):
+        ranked = CandidateSet(articles=self.all_articles)
+        score = rank_bias_entropy(ranked, k=3)
+        self.assertEqual(score, 1.3250112108241772)
```
mdekstrand, line 55: After switching to PyTest, you will be able to write this as:

```python
assert score == approx(1.325)
```

and it will be less brittle to slight FP rounding changes, and easier to read. (Unit test might have an approximate equality assertion, but PyTest definitely does).

mdekstrand (changes requested):
Sorry it took so long!

I think the test content looks good.

I have some comments on specifics of the rank-biasing metrics to correct the geometric model, make them consistent, and possibly reuse code.

The big thing, though: I hate to ask for this much change, but the rest of the tests use `pytest` instead of `unittest`, and I am hesitant for us to add a second testing framework to the repository. It shouldn't be too hard to change — lots of vscode multi-cursor editing — but consistency would be good IMO. PyTest also has good support for approximate floating-point tests.

mdekstrand:
The remaining test fix can be done after switching to PyTest, and then adding logic to a test fixture or a test marker to only run these tests when the data is available (since they require data), or modify them to no longer require that data.

mdekstrand:
@aisa6148 The rank weights are now documented: https://lkpy.lenskit.org/latest/api/metrics#rank-weights

src/poprox_recommender/evaluation/metrics/rbe.py
aisa6148: Changed it 0.85

src/poprox_recommender/evaluation/metrics/rbe.py
aisa6148: Please let me know if this was the expectation: I updated the lenskit version and used GeometricRankWeight to calculate weight for RBO and RBE.

src/poprox_recommender/evaluation/metrics/rbo.py
```diff
@@ -1,24 +1,34 @@
...
     https://dl.acm.org/doi/10.1145/1852102.1852106
     """
-    summands = []
-    for d in range(1, k + 1):
+
+    if p == 0:
+        raise Exception("RBO: p=0")
+
+    rank_weight = GeometricRankWeight(p)
+    weights = [rank_weight.weight(d) for d in range(1, k + 1)]
```
mdekstrand, line 19: `weight` takes an array, so can just write:

```python
weights = rank_weight.weight(np.arange(1, k+1))
```

Creating a list here will create a list of vectors, likely won't do what is wanted.

src/poprox_recommender/evaluation/metrics/rbo.py
```diff
@@ -1,24 +1,34 @@
...
+    total_weights = 0
+
+    for d, weight in enumerate(weights, start=1):
         set_a = set([a.article_id for a in recs_list_a.articles[:d]])
         set_b = set([b.article_id for b in recs_list_b.articles[:d]])
         overlap = len(set_a.intersection(set_b))
         agreement = overlap / d
-        summands.append(agreement * np.power(p, d))
+        sum += agreement * weight
+        total_weights += weight
```
mdekstrand, line 30: This is good for POPROX.

When we go to LensKit we should vectorize this.

tests/test_lip.py
```diff
@@ -0,0 +1,81 @@
...
+    reranked = CandidateSet(articles=all_articles[:10])
+    lip_score = least_item_promoted(reference, reranked, k=10)
+    assert lip_score == 0.0
+
+
+def test_mild_promotion_beyond_k(all_articles):
+    reference = CandidateSet(articles=all_articles[:5])
+    reranked = CandidateSet(articles=all_articles[2:5] + all_articles[0:2])
+    lip_score = least_item_promoted(reference, reranked, k=3)
+    assert pytest.approx(lip_score, 0.00001) == 0.4
```
mdekstrand, line 47: conventional use of approx goes the other way:

```python
assert lip_score == approx(0.4)
```

src/poprox_recommender/evaluation/metrics/lip.py
```diff
@@ -0,0 +1,18 @@
...
+
+def least_item_promoted(reference_article_set: CandidateSet, reranked_article_set: CandidateSet, k: int = 10) -> float:
+    reference_ids = [article.article_id for article in reference_article_set.articles]
+    reranked_ids = [article.article_id for article in reranked_article_set.articles]
+    reranked_position_map = {aid: i + 1 for i, aid in enumerate(reranked_ids)}
+    lip_rank = 0
+
+    for i, aid in enumerate(reference_ids):
+        rerank_pos = reranked_position_map.get(aid, k + 1)
+        if rerank_pos > k and rerank_pos > lip_rank:
```
mdekstrand, line 12: Why check for `rerank_pos > k`? Won't that only look at items that are reranked, but are not reranked into the top `k`? But I'll defer to @rburke2233 on the correctness of this metric.

(What is the reference for LIP?)

tests/test_lip.py
```diff
@@ -0,0 +1,81 @@
...
+    reference = CandidateSet(articles=all_articles[:5])
+    reranked = CandidateSet(
+        articles=[all_articles[1], all_articles[2], all_articles[6], all_articles[0], all_articles[4]]
+    )
+    lip_score = least_item_promoted(reference, reranked, k=3)
+    logger.info("test_partial_overlap %s", lip_score)
+    assert lip_score == 0.4
+
+
+def test_empty_reference(all_articles):
```
mdekstrand, line 68: If the reference list is empty, where is the reranker getting its articles? Probably a fine test, but maybe the metric should be `NaN` in this case? (You can't measure LIP when there are no items to promote.)

src/poprox_recommender/evaluation/metrics/rbo.py
```diff
@@ -1,24 +1,34 @@
...
     Webber, William, Alistair Moffat, and Justin Zobel. "A similarity measure for indefinite rankings."
     ACM Transactions on Information Systems (TOIS) 28.4 (2010): 20.
 
     https://dl.acm.org/doi/10.1145/1852102.1852106
     """
-    summands = []
-    for d in range(1, k + 1):
+
+    if p == 0:
+        raise Exception("RBO: p=0")
```
mdekstrand, line 16: `ValueError` would be the conventional exception here. Or you can do nothing — `GeometricRankWeight` validates its inputs.

src/poprox_recommender/evaluation/metrics/rbe.py
```diff
@@ -0,0 +1,36 @@
...
+
+mind_data = MindData()
+
+
+def rank_bias_entropy(final_recs: CandidateSet, k: int, d: float = 0.85):
+    top_k_articles = final_recs.articles[:k]
+    weighted_counts = defaultdict(float)
+
+    rank_weight = GeometricRankWeight(d)
+    weights = [rank_weight.weight(rank) for rank in range(1, k + 1)]
```
mdekstrand, line 17: Same commment as in RBO — pass an array of ranks, `weight` is vectorized.

mdekstrand (changes requested):
Pursuant to the discussion in Slack, let's get the metrics that work correctly on single lists merged, and remove the rest in favor of LensKit-first.

Metrics that can be kept AFAIK:

- ILS
- RBE
- LIP (I've got a couple questions on correctness, though)
- RBO (needs adjustment of how it uses weights)

karlhigley:
From the test runs, it looks like many of the metric definitions are creating their own `MindData` object and failing to find the actual MIND data files. Two questions:
* Should we be passing a `MindData` object in as a parameter where needed to make sure we're doing lookups against the same data we're evaluating on? Seems like we should, but maybe I'm missing something.
* What's our strategy for testing things that require MIND data? I think we use it in some other tests but I forget how that works.

mdekstrand:
@karlhigley yes, we should pass in minddata (or a better abstraction) to the metrics.

On tests: we have some logic elsewhere to `xfail` tests that need MIND data, after checking an environment variable. This can be automated by accepting the MIND data as a fixture.

src/poprox_recommender/evaluation/metrics/rbe.py
```diff
@@ -0,0 +1,35 @@
...
+    top_k_articles = final_recs.articles[:k]
+    weighted_counts = defaultdict(float)
+
+    rank_weight = GeometricRankWeight(d)
+    weights = rank_weight.weight(np.arange(1, k + 1))
+
+    for rank, (article, weight) in enumerate(zip(top_k_articles, weights), start=1):
+        mentions = article.mentions
+        if not mentions:
+            article_details = mind_data.lookup_article(uuid=article.article_id)
```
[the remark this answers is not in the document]
mdekstrand: yes, that's a good point. (we should only construct one MIND data.)

mdekstrand:
On only constructing one MIND data: this will require a small extra jump due to parallel measurement.

1. In the main process, we should put the MIND data into Ray with `ray.put`, to get an object ref to it.
2. Pass this ref along with the other inputs to the worker function(s). Ray will automatically fetch the object ref before calling the worker.
3. In the worker, pass the MIND data to the metrics that need it.

karlhigley:
🙌🏻

aisa6148:
> 🙌🏻

I am sorry it took so long!

PR #183, opened 2025-04-16
Refactor generation and evaluation to use Ray

mdekstrand:
This refactors our batch-inference and evaluation code to use Ray instead of ipyparallel; even the ipyparallel people suggest Ray these days. It also tweaks our eval dependency setup to use a UV group instead of extra, making it easier to make sure we have the parallelism bits installed.

Key changes:

- Batch-inference and evaluation now use Ray to parallelize instead of ipyparallel.
- Parallelism now uses LensKit's parallel processing configuration, so we configure with `LK_NUM_PROCS` instead of `POPROX_NUM_PROCS`. LensKit also automatically sets up Ray workers to limit per-process parallelism to avoid oversubscription.
- Remove a bunch of parallel logging configuration that is now handled by LensKit's Ray support.
- Make the embedding writer its own Ray actor, so we don't need to write to temporary files and then deduplicate.

This is Part 1 of a two-part refactor of offline evaluation. Part 2 will separate pipelines into their own evaluation jobs, so we don't have to re-evaluate the entire set of pipelines just because one changed.

Adding Ray as a dependency and using it for our batch inference also begins paving the way for later using ray-tune to do hyperparameter tuning.


mdekstrand:
It is possible that trying to run this on a system with a GPU that is too small will cause it to fail to start, because it cannot start enough actors. I'm not sure what the prettiest solution to make that fail gracefully is.

karlhigley (approved):

PR #189, opened 2025-05-01
Refactor evaluation to expose separate tasks for each pipeline

mdekstrand:
This PR updates the evaluation to expose separate tasks for each pipeline and output more complete recommendation data.

- Take pipeline name as an input to the evaluation scripts, so we can run one pipeline without rerunning all (making it much less expensive to compare a new pipeline against the ones already in the repository).
- Write full recommendation output to `recommendations.ndjson.zst` so we can get articles, headlines, etc.
- Abstract recommendation data extraction and writing, producing each output in a separate actor (so output is concurrent with inference). Tabular Parquet, JSON, and embedding collection are all just "recommendation writers".
- Use Ray tasks instead of actors for generating recommendations, simplifying parallelism configuration.

### DIFF
--- a/dvc.yaml
+++ b/dvc.yaml
@@ -15,7 +15,7 @@ stages:
   recommend-mind-val:
     foreach: ${pipelines}
     do:
-      cmd: python -m poprox_recommender.evaluation.generate -M MINDlarge_dev -o outputs/mind-val ${item}
+      cmd: python -m poprox_recommender.evaluation.generate -M MINDlarge_dev -o outputs/mind-val/${item} ${item}
       deps:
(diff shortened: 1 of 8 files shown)


mdekstrand:
This is basically ready for review — all the interesting bits are in place. The two things before merge:

- re-run remaining output to get metrics
- fix the shared-data thing so we aren't publishing MIND content

mdekstrand:
Right now, this Ray actor design bottlenecks on the output actors. Trying to debug that, but our current net speed is definitely reduced from optimal.

karlhigley (approved):

PR #192, opened 2025-05-28
Fix score overwriting issues caused by mutability with simple copy-on-write 

karlhigley:
This also corrects an unintended behavior caused by checking interacted articles instead of the click history, which used to be equivalent but no longer is because we now have virtual topic-based "articles" and articles that people have given 👍🏻 / 👎🏻  feedback on.

### DIFF
--- a/src/poprox_recommender/components/embedders/user.py
+++ b/src/poprox_recommender/components/embedders/user.py
@@ -32,7 +32,7 @@ class NRMSUserEmbedder(Component):
 
     @torch_inference
     def __call__(self, clicked_articles: CandidateSet, interest_profile: InterestProfile) -> InterestProfile:
-        if len(clicked_articles.articles) == 0:
+        if len(interest_profile.click_history) == 0:
             interest_profile.embedding = None
         else:
             embedding_lookup = {}


Sris-Ty (approved):

src/poprox_recommender/components/joiners/score.py
```diff
@@ -17,15 +17,17 @@ def __call__(self, candidates1: CandidateSet, candidates2: CandidateSet) -> Cand
...
+            for article, score in zip(candidates1.articles, candidates1.scores):
+                article_id = article.article_id
+                combined_score[article_id] += score
+                combined_article[article_id] = article
+
+        if candidates2.scores is not None:
+            for article, score in zip(candidates2.articles, candidates2.scores):
+                article_id = article.article_id
+                combined_score[article_id] += score
+                combined_article[article_id] = article
```
mdekstrand, line 30: This fixes the problem in #193.

src/poprox_recommender/components/scorers/article.py
```diff
@@ -13,10 +13,12 @@ def __call__(self, candidate_articles: CandidateSet, interest_profile: InterestP
         candidate_embeddings = candidate_articles.embeddings
         user_embedding = interest_profile.embedding
 
+        with_scores = candidate_articles.model_copy()
```
mdekstrand, line 16: with #193, this will no longer be needed — do we want to remove there? here? merge both together? leave it here and do an extra (cheap) copy?

mdekstrand (approved):

PR #193, opened 2025-05-30
Use LensKit hooks to proactively shallow-copy Pydantic models

mdekstrand:
This addresses the model-data-overwriting problem by using a LensKit pipeline hook to proactively make a shallow copy of any Pydantic model that is routed to a component as one of its inputs.


karlhigley (approved):

mdekstrand:
@karlhigley fixed it before seeing your comment, but yes, it is. Change I just pushed is identical to the `ScoreFusion` change in #192.

PR #196, opened 2025-06-05
Train NRMS on a subset

sushobhan2024:
Train NRMS on a subset with arg `--subset=n` for debugging. Takes 1.25 minutes on `cruncher` for `n=1000`. 
It now saves output for subset training in a separate folder.


src/poprox_recommender/training/dataset.py
```diff
@@ -37,6 +37,9 @@ def __init__(self, args, behaviors_path, news_path):
         # if empty, padding title token by zeros
         self.padding = {k: v for k, v in padding_all.items() if k in [args.dataset_attributes]}
 
+        if hasattr(self.args, "subset") and self.args.subset is not None:
```
mdekstrand, line 40: Shouldn't need `hasattr` here.

mdekstrand (approved):

src/poprox_recommender/training/dataset.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, args, behaviors_path, news_path):
         # if empty, padding title token by zeros
         self.padding = {k: v for k, v in padding_all.items() if k in [args.dataset_attributes]}
 
-        if hasattr(self.args, "subset") and self.args.subset is not None:
+        if self.args.subset and self.args.subset is not None:
```
mdekstrand, line 40: Just need `self.args.subset is not None`.

src/poprox_recommender/training/train.py
```diff
@@ -144,11 +144,20 @@ def save_model(model):
...
+
     lc = LoggingConfig()
     if args.verbose:
         lc.set_verbose(True)
     lc.apply()
 
     torch.cuda.empty_cache()
 
-    train(device, args.load_checkpoint, args)
+    train(device, args.load_checkpoint, args, args.output_dir)
```
mdekstrand, line 163: Since we're passing `args`, why not just have `train` look at it instead of passing `args.output_dir` separately?

src/poprox_recommender/training/train.py
```diff
@@ -144,11 +144,20 @@ def save_model(model):
     parser.add_argument("--num_words_title", type=float, default=30)
 
     args = parser.parse_args()
+
+    # set a seperate output directory for subset
+    if args.subset is not None:
+        args.output_dir = root / f"models/nrms-mind-subset-{args.subset}"
+    else:
+        args.output_dir = MODEL_DIR
```
mdekstrand, line 152: We should either just use an `output_dir` variable (don't put it on `args`), and pass it below (and ignore my comment below); or we should add a `--output-dir` argument in `add_argument` above, and only do this setting if `args.output_dir is None` (i.e., if they don't specify an output dir, we ovewrite it).

mdekstrand: The reason is that `args` should be the command line arguments. So if `--output-dir` is a command-line argument, then we should use `args.output_dir`; if it is not, then we should use a separate variable, to avoid surprising future readers.

mdekstrand (approved):
One thing to clean up / make consistent on output directories and command-line arguments and then we are good to merge.

src/poprox_recommender/training/train.py
```diff
@@ -146,8 +147,9 @@ def save_model(model, output_dir):
     args = parser.parse_args()
 
     # set a seperate output directory for subset
-    if args.subset is not None:
-        args.output_dir = root / f"models/nrms-mind-subset-{args.subset}"
+    if args.output_dir is None:
+        if args.subset is not None:
+            args.output_dir = root / f"models/nrms-mind-subset-{args.subset}"
     else:
         args.output_dir = MODEL_DIR
```
mdekstrand, line 154: this should indent a level

mdekstrand (approved):

PR #202, opened 2025-06-12
Add measurement and retrain NRMS

mdekstrand:
This adds better training measurement support and retrains the NRMS model.

- [x] Add `lenskit.toml` to configure power metering.
- [x] Wrap training in a `Task` to record resource consumption and save it with model checkpoints.
- [x] Upgrade dependency to published version with the power and configuration logic.
- [x] Finish running model training.


mdekstrand:
Only reason I see that might block merge is if we want the new eval results first.

karlhigley (approved):

PR #206, opened 2025-06-23
Enable multi-GPU training with Ray Train

karlhigley:
The changes to our training script are surprisingly minor but look bigger than they are due to whitespace changes.

### DIFF
--- a/src/poprox_recommender/training/train.py
+++ b/src/poprox_recommender/training/train.py
@@ -3,9 +3,12 @@ from os import fspath
 
 import lenskit
 import lenskit.config
+import ray.train.huggingface.transformers
 import torch
 from humanize import metric
 from lenskit.logging import LoggingConfig, Task, friendly_duration, get_logger
+from ray.train import ScalingConfig
+from ray.train.torch import TorchTrainer
 from safetensors.torch import load_file, save_file
 from transformers import Trainer, TrainingArguments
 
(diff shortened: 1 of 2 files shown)


src/poprox_recommender/training/train.py
```diff
@@ -168,7 +165,19 @@ def save_model(model, output_dir):
 
     with Task("train NRMS") as task:
         task.save_to_file(args.output_dir / "task.json")
-        train(device, args)
+        train_func = train(device, args)
```
mdekstrand, line 168: If we manually enable the cluster with LensKit's [`init_cluster`](https://lkpy.lenskit.org/latest/api/lenskit.parallel.ray.html#lenskit.parallel.ray.init_cluster), using `init_cluster(global_logging=True)`, it will wire up the subprocess logging elements. Ideal place for this would be after `lenskit.configure` and before the `with` block.

src/poprox_recommender/training/train.py
```diff
@@ -20,90 +23,84 @@
...
-        if torch.isnan(tensor).any():
-            raise ValueError(f"NaN in {name}: {tensor}")
+        # Prepare Transformers trainer
+        trainer = ray.train.huggingface.transformers.prepare_trainer(trainer)
 
-    for param_name, param_value in model.named_parameters():
-        if param_value.grad is not None:
-            check_nan(param_value.grad, f"gradient[{param_name}]")
-    """
+        trainer.train()
```
mdekstrand, line 101: If we wrap this in a `Task`, we can get subprocess task tracking too:

```python
with Task("training worker", subprocess=True) as task:
    trainer.train()
send_task(task)
```

This is necessary for the CPU time statistics to be accurate, but _not_ for the power statistics.

(The ergonomics of this could maybe use some work. Perhaps `send_task` should be automatic. But with current code, that's how you wire up a subprocess task meter.)

mdekstrand (commented):
Left a couple comments about logging and resource training.

Do we want to conditionalize this logic? Being able to train without Ray seems useful for debugging and profiling.

karlhigley:
Closing this since I'm probably the only person it matters to at this point

PR #211, opened 2025-07-03
generating image embedding with clip

Mahamudul42:
generating image embedding with clip


karlhigley:
Closing because this moved to a fork of the repo and has already been deployed as a separate endpoint

PR #213, opened 2025-07-10
image personalization using clip embedding!

Mahamudul42:
image personalization using clip embedding!

### DIFF
--- a/src/poprox_recommender/components/filters/image_selector_clip.py
+++ b/src/poprox_recommender/components/filters/image_selector_clip.py
@@ -1,0 +1,132 @@
+import logging
+from io import BytesIO
+
+import requests
+import torch
+from PIL import Image as PILImage
+from torchvision import transforms
+from transformers import CLIPModel, CLIPProcessor
+
+from poprox_concepts.domain import CandidateSet, InterestProfile, RecommendationList
+
+logger = logging.getLogger(__name__)
+
+
+class ClipImageSelector:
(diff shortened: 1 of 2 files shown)


karlhigley (commented):
The code looks great! There are a few things I want to sort out before we merge, which are more about demoing and deployment than the code itself:
* We should set up with a fork of the recommender repo you can use to deploy a separate endpoint that serves recs for an experiment. This should be pretty quick and we can work on it together next time we meet. Once the new fork exists, we can merge this code into it as is.
* The recommender pipeline here is a reasonable place to start, but it's based on a different recommender than the default (`nrms_topic_scores`) that we're using to send daily newsletters. I remember your advisor being a bit confused by receiving different article recs before, which ended up being a distraction from the image selection you wanted to highlight. You can put the `ClipImageSelector` on the end of any pipeline the same way you have here though, so adjusting this should be mostly straightforward copy/paste.
* The `__init__()` method is loading a pre-trained model by name, which works fine for local testing but doesn't work as well on AWS since it downloads the model on the fly. That can cause endpoint timeouts, which is the main reason I hesitate to merge this into the main fork right away. We should bake the CLIP model into the container, which is mostly a logistical thing that won't affect the code here very much.

karlhigley:
(Looks like the tests are failing due to a missing `torchvision` dependency. Should be an easy fix, but let's get your fork set up first.)

PR #216, opened 2025-07-21
Remove unnecessary direct dependencies

mdekstrand:
This removes the unnecessary direct dependencies from `pyproject.toml`, that I think are leftovers from previous installation methods for LensKit. In particular, it drops the direct `numpy` dep, which will let us use NumPy 2.0, and hopefully simplify the manual site build process.

This also removes the `.python-version` file, relying on the `requires-python` setting as the single source of truth on the required Python version, and updates our `setup-uv` actions.


karlhigley (approved):

karlhigley:
Looks like it's trying to fetch `hatchling` from LensKit PyPI but otherwise looks good

mdekstrand:
@karlhigley I have no idea why it was doing that, but after some `uv` updates it seems to no longer have that problem. PR now has an update to our usage of `setup-uv`, which also then simplifies subsequent commands in several actions.

PR #217, opened 2025-07-21
Remove recommenders that have moved to forks

karlhigley:
These recommenders aren't going away but going forward they'll be served by separate endpoints (and not by the default recommender endpoint.)

We have more recommenders (esp. around topics and feedback) that should also move to a separate fork/endpoint, but we're not quite ready to remove those here yet.


mdekstrand (commented):
We also need to remove these models from the `vars.pipelines` definition in `dvc.yaml`.

PR #218, opened 2025-07-21
Correctly resolve model path for news encoder

mdekstrand:
This might fix Aishwarya's other problem. I think the old code was re-fetching pretrained models from HuggingFace and we didn't realize it.


mdekstrand:
I wonder if our model retrain was also completely wrong, using a different distilbert than we baked in?

karlhigley (approved):
Puzzled why this would make a difference but the change looks fine

PR #224, opened 2025-07-29
refactoring the user embedder into single and multi vector

Sris-Ty:
FM style NRMS requires multiple vector user ember which is almost similar to the NRMS user embedder except for the last additive layer which leads to single vector user embedder and multi vector user embedder.


karlhigley:
Closing old/inactive PRs to clarify what's in progress

PR #226, opened 2025-07-29
Use pre-computed embeddings in the image selector

karlhigley:
This just refactors the component to move embedding outside of it and use a lookup table of pre-computed embeddings


karlhigley:
Closing because this is pointed at the wrong fork

karlhigley:
The real PR is over here: https://github.com/Mahamudul42/poprox-recommender-clip/pull/3

PR #228, opened 2025-07-30
Least Item Promoted Fix

aisa6148:
Current LIP implementation would always return 0 since the reference set is terminated at top k
TODO: Test cases to be fixed accordingly


karlhigley:
I think this has been addressed a different way (by getting longer lists of items beyond top-k)

PR #294, opened 2026-03-13
Replace RecLoader with direct streaming from compressed NDJSON

sushobhan2024:
Previously, measurement loaded Parquet into DuckDB with workers looking up slates by ID. This doesn't scale to compressed JSON which can't be efficiently loaded into DuckDB. This PR replaces `RecLoader` with direct streaming of the `.ndjson.zst` file line by line. The driver batches raw bytes and sends them to Ray workers, which parse each line into `OfflineRecommendations`. 

Key changes:
- Remove `RecLoader` class
- `recommendation_eval_results` now takes a `Path` to the `.ndjson.zst` file
- `measure_rec_metrics` accepts `OfflineRecResults` instead of `pd.DataFrame`
- Flatten `list[ImpressedSection]` at metric time
- Progress bar no longer shows total count, no longer countable without reading the full file
- Update measure stage deps in `dvc.yaml` from recommendations.parquet
  to `recommendations.ndjson.zst` to match new streaming measurement path


sushobhan2024:
The old `RecLoader` had a bug that inserted Parquet data twice, that duplicated every recommendation. This inflated all ranking metrics (RBP, NDCG, MRR) by approximately 2x. The changes made with `NDJSON` streaming path correctly reads each recommendation once, which should be the correct metric results.

mdekstrand (approved):
We need to fix the tests - possibly by adding a group to the UV setup install - but otherwise good.

PR #302, opened 2026-04-06
Add a sectioned recommender with non-personalized top news section

karlhigley:


src/poprox_recommender/recommenders/configurations/nrms_sections_nptn.py
```diff
@@ -0,0 +1,323 @@
+from uuid import UUID
```
karlhigley, line 1: It's hard to tell from the diff, but this file is _exactly_ the same as the `nrms_sections` pipeline, except in one place I'll call out below

src/poprox_recommender/recommenders/configurations/nrms_sections_nptn.py
```diff
@@ -0,0 +1,323 @@
...
+        "ts_candidates",
+        TopGeneralNewsCandidates,
+        candidate_articles=fusion,
+        article_packages=i_packages,
+    )
+
+    ts_topk = builder.add_component("ts_topk", TopkRanker, TopkConfig(num_slots=3), candidate_articles=ts_candidates)
+
+    ts_config = AddSectionConfig(title="Top Stories", personalized=False)
+    ts_sections = builder.add_component("top_stories", AddSection, ts_config, new_section=ts_topk)
```
karlhigley, line 218: Since this section isn't personalized, we don't need any of the topic prefs filtering shenanigans that were here before. Now we just create a candidate pool of the top ten General News articles in order but without relevance scores and pass it to `TopkRanker`, which will maintain the original order if it doesn't find scores.

karlhigley:
Since this has no impact on the main/default recommender, I'm going to go ahead and ship it in the interest of iteration speed and testing with the live system via test newsletters and a preview experience.

kluver (commented):
lgtm. let's get this pipeline in testing.

PR #304, opened 2026-04-07
Add a pipeline with a disliked topic as the middle topical section

karlhigley:
This deliberately puts a topic we know the user doesn't like (based on their topic prefs) in the middle section slot. We still try to pick the most relevant articles from that topic.

A potential experiment using this kind of treatment might be a way to look at the impact of visibly ignoring information POPROX collects from users on their sense of control and satisfaction.

### DIFF
--- a/src/poprox_recommender/recommenders/configurations/nrms_sections_disliked_topic.py
+++ b/src/poprox_recommender/recommenders/configurations/nrms_sections_disliked_topic.py
@@ -1,0 +1,340 @@
+from uuid import UUID
+
+from lenskit.pipeline import PipelineBuilder
+
+from poprox_concepts.domain import ArticlePackage, CandidateSet, InterestProfile
+from poprox_recommender.components.embedders import NRMSArticleEmbedder
+from poprox_recommender.components.embedders.article import NRMSArticleEmbedderConfig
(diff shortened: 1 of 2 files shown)


src/poprox_recommender/recommenders/configurations/nrms_sections_disliked_topic.py
```diff
@@ -0,0 +1,340 @@
+from uuid import UUID
```
karlhigley, line 1: This whole file is literally exactly the same as the `nrms_sections` pipeline, except for one thing called out below

src/poprox_recommender/recommenders/configurations/nrms_sections_disliked_topic.py
```diff
@@ -0,0 +1,340 @@
...
+    )
+
+    topic2_candidates = builder.add_component(
+        "topic2_candidates",
+        TopicalCandidates,
+        TopicalCandidatesConfig(min_candidates=3),
+        candidate_set=topic2_deduped,
+        interest_profile=i_profile,
+        sections=topic1_sections,
+        descending=False,
```
karlhigley, line 270: The second topical section sorts topics by _ascending_ preference instead of _descending_, putting the most disliked topics first in the list of candidate topics (but just for this one section)

PR #305, opened 2026-04-08
Add a sectioned recommender that only recommends sports

karlhigley:
Minor refactoring of the existing filter based on topic prefs, a new component to include or exclude articles that mention specific entities, and some quick pipeline changes

### DIFF
--- a/src/poprox_recommender/recommenders/configurations/nrms_sections_only_sports.py
+++ b/src/poprox_recommender/recommenders/configurations/nrms_sections_only_sports.py
@@ -1,0 +1,340 @@
+from uuid import UUID
+
+from lenskit.pipeline import PipelineBuilder
+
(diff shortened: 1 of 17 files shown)


src/poprox_recommender/recommenders/configurations/nrms_sections_only_sports.py
```diff
@@ -0,0 +1,341 @@
...
+        candidates2=explicit_fusion,
+    )
+
+    # Filter out stories that don't mention sports
+    f_mentioned = builder.add_component(
+        "sports-filter",
+        MentionedEntitiesFilter,
+        MentionedEntitiesConfig(entities=set(["Sports"]), entity_type="topic"),
+        candidates=fusion,
+    )
```
karlhigley, line 213: Here we filter out anything that doesn't mention Sports (at any relevance level), and below we omit the filter for In Other News that removes articles in topics that were used for topical sections

PR #306, opened 2026-04-08
Add a recommender that selects topical sections before top stories

karlhigley:


src/poprox_recommender/components/sections/combine.py
```diff
@@ -33,3 +34,28 @@ def __call__(
             existing_sections.append(new_section)
 
         return existing_sections
+
+
+AddSection: TypeAlias = AppendSection
+
+
+class PrependSection(Component):
```
karlhigley, line 42: To choose things out of order, we need a way to insert sections earlier in the list. For now, inserting them at the beginning is enough, but an alternate version of this component could insert them at a specific position

src/poprox_recommender/recommenders/configurations/nrms_sections_reordered.py
```diff
@@ -0,0 +1,343 @@
+from uuid import UUID
```
karlhigley, line 1: As usual, this is mostly the same as `nrms_sections`, but I'll call out the differences below

src/poprox_recommender/recommenders/configurations/nrms_sections_reordered.py
```diff
@@ -0,0 +1,343 @@
...
+    # Combine click and explicit feedback -> all preference
+    fusion = builder.add_component(
+        "fusion",
+        ScoreFusion,
+        {"combiner": "avg", "weight1": 1, "weight2": 2},
+        candidates1=n_scorer,
+        candidates2=explicit_fusion,
+    )
+
+    # Topical Sections
```
karlhigley, line 207: The topical section selection is basically the same, except now there are no previous sections to take into account so we don't need a filter to de-duplicate items that are already in the newsletter

src/poprox_recommender/recommenders/configurations/nrms_sections_reordered.py
```diff
@@ -0,0 +1,343 @@
...
+    topic3_sections = builder.add_component(
+        "topic3_sections",
+        AppendSection,
+        AddSectionConfig(personalized=True),
+        new_section=topic3_topk,
+        existing_sections=topic2_sections,
+    )
+
+    # Top Stories
+    yts_deduped = builder.add_component("yts_deduped", DuplicateFilter, candidate=fusion, sections=topic3_sections)
```
karlhigley, line 275: Here we remove items that are already in the topical sections from the pool of available candidates

src/poprox_recommender/recommenders/configurations/nrms_sections_reordered.py
```diff
@@ -0,0 +1,343 @@
...
+    )
+
+    yts_config = AddSectionConfig(title="Your Top Stories", personalized=True)
+    yts_sections = builder.add_component(
+        "top_stories",
+        PrependSection,
+        yts_config,
+        new_section=yts_fill,
+        existing_sections=topic3_sections,
+    )
```
karlhigley, line 309: Once we've built the Top Stories section, we prepend it to the list of sections so that it appears before the topical sections (even though it was constructed after)

src/poprox_recommender/recommenders/configurations/nrms_sections_reordered.py
```diff
@@ -0,0 +1,343 @@
...
+
+    # In Other News section
+    ion_deduped = builder.add_component("ion_deduped", DuplicateFilter, candidate=fusion, sections=yts_sections)
```
karlhigley, line 312: Then the In Other News section is built based on the list of sections we just prepended to

src/poprox_recommender/recommenders/configurations/nrms_sections_reordered.py
```diff
@@ -0,0 +1,343 @@
...
+
+    ion_narrowed = builder.add_component(
+        "ion_narrowed",
+        PreviousSectionsFilter,
+        candidate=ion_deduped,
+        article_packages=i_packages,
+        sections=topic3_sections,
```
karlhigley, line 319: The `PreviousSectionsFilter` removes articles about entities that already got a dedicated section, but the Your Top Stories section doesn't matter here because it doesn't have a seed entity that needs to be filtered out

src/poprox_recommender/recommenders/configurations/nrms_sections_reordered.py
```diff
@@ -0,0 +1,343 @@
...
+    )
+
+    ion_fill = builder.add_component(
+        "ion_fill", FillRecs, FillConfig(num_slots=3), recs1=ion_topk_filtered, recs2=ion_topk_unfiltered
+    )
+
+    ion_config = AddSectionConfig(title="In Other News", personalized=True)
+    builder.add_component(
+        "recommender", AppendSection, ion_config, new_section=ion_fill, existing_sections=yts_sections
+    )
```
karlhigley, line 343: Here it does matter that we use the sections from after prepending YTS though so that we end up with the full list of one Top Stories section, three topical sections, and one In Other News section

PR #309, opened 2026-05-08
Section-aware RBP metric with browsing model weights

sushobhan2024:
- Added `section_browsing_model.py` with `section_browsing_weights` function. Per-item visit probabilities are computed using abandon, skip, and continue parameters: `0.10`, `0.20`, and `0.70` respectively. 
- Added `section_aware_RBP` to `measures_rec_metrics` 
- Added tests for `section_browsing_weights`

### DIFF
--- a/tests/test_section_browsing_model.py
+++ b/tests/test_section_browsing_model.py
@@ -1,0 +1,77 @@
+from uuid import uuid4
+
+import numpy as np
+import pytest
+
+from poprox_concepts.domain import Article, ImpressedSection, Impression
+from poprox_recommender.evaluation.measure.section_browsing_model import section_browsing_weights
+
+
+def make_section(n_articles: int) -> ImpressedSection:
+    impressions = [Impression(article=Article(article_id=uuid4(), headline=f"Article {i}")) for i in range(n_articles)]
+    return ImpressedSection(impressions=impressions)
+
+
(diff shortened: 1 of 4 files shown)


tests/test_section_browsing_model.py
```diff
@@ -0,0 +1,77 @@
...
+            0.256118,  # Section 2
+            0.390293,
+            0.273205,
+            0.191243,  # Section 3
+            0.291432,
+            0.204002,
+            0.142801,  # Section 4
+            0.217612,
+            0.152328,
+            0.10663,  # Section 5
```
mdekstrand, line 45: Do we want to normalize to max to 1, like we did in the [mockup](https://docs.google.com/spreadsheets/d/1ssPWQKgABQ_Xcxhu1-OMgBjxqB9DgI6h/edit?gid=404127243#gid=404127243)?

mdekstrand (approved):

tests/test_section_browsing_model.py
sushobhan2024: The mockup note says 'recommend using probabilities directly as our weights' — should we follow that or switch to normalized?

mdekstrand: we can stick with the probabilities.

PR #312, opened 2026-05-21
Upgrade to LensKit 2026

mdekstrand:
This upgrades POPROX to use LensKit 2026, which paves the way to enable pipeline visualization.


mdekstrand:
Ok good, the tests pass. The Docker test failure is due to the Git dependency, and our Docker image not having a Rust compiler. Once the prebuilt wheel is available, it should start passing. But I wanted to make sure the relevant change (lenskit/lkpy#1102) made our tests pass before considering the bug (lenskit/lkpy#1101) fixed and merging it into LensKit.

PR #313, opened 2026-05-25
Export pipelines to JSON and Mermaid

mdekstrand:
This uses the new LensKit abilities to export pipeline configurations to JSON and Mermaid.

It also brings some small changes to the filter pipeline (did @Sris-Ty build this initially?) to align configuration with usage, which comes up when working with the pipeline configurations for visualization.

### DIFF
--- a/pipelines/nrms.json
+++ b/pipelines/nrms.json
@@ -1,0 +1,122 @@
+{
+  "options": null,
+  "meta": {
+    "name": "nrms",
+    "version": "0.0.1",
+    "hash": "c75d89a98cbbe346f72e00183712f5b027231222edbfdafb13c55e960da2258b"
+  },
+  "inputs": [
+    {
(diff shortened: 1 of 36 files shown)


Sris-Ty:
Karl and I initially wrote the nrms_feedback pipeline. Later, for an internal experiment, Karl implemented the nrms_feedback_filter code, which we then tested. The idea was to exclude candidate articles whose topics largely fall within the lower end of a user's topical preferences.

Sris-Ty (approved):
I ran my local testing script to generate recommendations using nrms_article_feedback and compared the results with nrms_topic_scores for a pseudo user profile under different combinations of article-level feedback, clicks, and topical preferences. So far, it does not seem to be breaking anything.

mdekstrand:
Does anyone know why this test is failing?

